PR #14, opened 2019-03-06
8 reporting

### LINKED ISSUE #8: Generate reporting statistics
This supports ONSdigital/sdg-indicators#2851

Alongside individual indicator information we should output statistics across indicators. Propose a whole route of `/stats/`. Starting with `reporting`. For the site we need quite detailed information so that they can build the reporting page.

Initial draft of how it could look is here: https://gist.github.com/dougmet/98a7be9b5c92e0ab0bfe6aee141bf6fa

dougmet:
This PR adds reporting status computation.

* Creates a new `stats` route in the output with `stats/reporting.json` the output
* Makes a more generic `Schema` class which loads from the prose config but opens up future possibilities of other formats.
* Makes use of the `reporting_status` field in the schema to get possible statuses
* Added 0.3.0 changelog as this is definitely a minor version increase.

_Possible_ breaking change:
This shouldn't break anything for anyone but I would advise checking their `_prose.yml` file in their data repo to make sure it reflects what is in their metadata. If a status is in the prose file it will be assumed it is used. If you don't want to use it, remove it.

Translation keys are expected in `_prose.yml` but defaults are present for:
* `complete` -> `status.reported_online`
* `inprogress` -> `status.statistics_in_progress`
* `notstarted` -> `status.exploring_data_sources`
* `notapplicable` -> `status.not_applicable`

which will be used if a translation key is not found.

Fixes #8


dougmet:
Have some Python 3.4 fixing to do. And forgot to add notapplicable.

dougmet:
The main Python 3.4 issue (apart from others that I've fixed) is that the pandas version (0.19) doesn't have `DataFrame.agg`. I will try to code around it but I'm seriously considering dropping 3.4 support as it's becoming a headache. It doesn't affect any current builds and would only impact someone who is running this on pretty old infrastructure. pandas 0.20.1 came out May 2017. What do you think @brockfanning ? 

I'll see if I can code around it for now.

dougmet:
OK 3.4 lives to fight another day. I've got my eye on it though 👀

dougmet:
More importantly I don't think I've fully accounted for translations. Lots to do.

brockfanning:
@dougmet Regarding translation, my hope would be that we could use dot-delimited "translation keys" (like from the sdg-translations project) instead of human-readable labels. For example, instead of "Reported online" it would be "status.reported_online", corresponding to the keys in [this file](https://github.com/open-sdg/sdg-translations/blob/master/translations/en/status.yml). One snag would be that would couldn't replace the Prose "label" with those, because then the field would show up as "status.reported_online" on Prose.io. But maybe we could add a new property in the _prose.yml schema, like "translation_key"?

dougmet:
Yes I think something like that could work. Although it would be simpler if the reporting_status value (`complete` etc) was just the translation key -- feels like we're double-keyed. Of course that means either changing everyone's metadata or the keys. Neither ideal.

I don't know if prose would just ignore a "translation" element or if it would error. That could be a natural place to map the status to the key without being too ugly. Worth a go.

brockfanning:
I did worry when I was making the sdg-translations project that chunking the translations into groups like that was arbitrary and unnecessary. Like you say it could just be "reported_online", or if we really need specificity, "status_reported_online". It's slightly useful in some cases, like in "global_indicators", where we can look up a particular indicator by its id ("global_goal.1"). But of course that could just as easily be "global_goals_1". You're right it would be a significant revamp at this point, but that doesn't mean we shouldn't consider it. Just need to decide if that grouping approach is useful or not.

dougmet:
I quite like the structure of `sdg-translations`. If I could change one thing it would be that the key for `complete` would be `status.complete` rather than `status.reported_online` but these kinds of things are inevitable I guess. I'm fairly happy with how I've worked around it in this PR.

The current output looks like this gist:
https://gist.github.com/dougmet/cdc5352cf0db87baf58d661d3b742919

dougmet:
Currently breaks if countries don't have the `translation_key` field in their `_prose.yml` config. For backwards compatibility it might be prudent to have some defaults.

dougmet:
Defaults have been added. I think this is ready to merge. There's no breaking changes now, beyond testing the prose file has all of the statuses that they want.

dougmet:
Oh that’s a good idea. If it’s quick I’ll add it otherwise I’ll add an issue.

Issue #4, opened 2019-02-04
Redesign backend with OO


dougmet:
Copied through from ONSdigital/sdg-build#9

This package has grown from disconnected scripts into something fairly reusable and testable. However, it's probably time to think about the design pattern of how data is stored on the backend. Here are some core classes that I think we'll want:

indicator: Hold data and metadata for an indicator. It will have write methods
site: Less certain on this. Some combination of the path functions and schema. Perhaps this is a holding class for indicator objects.

Post ideas here. Main requirement is that this is a back-end change and shouldn't break anything for sdg-data.

dougmet:
As noted from #9 need to think about how the internal data model serves sdmx as well as our immediate needs.

dougmet:
An early case of this is being tested with the `Schema` class in #14 . The idea here is that it can load from multiple schema files (only prose for now), store data internally in a consistent manner (a big dictionary for now) and write out to different formats (currently back to prose-style).

LucyGwilliamAdmin:
@brockfanning is this complete?

(sorry for all the notifications - just going through and seeing which ones can be closed)

brockfanning:
Yes I think we can close this one. There is still more we can do, but we can tackle those in new issues.

Issue #15, opened 2019-03-11
Write to SDMX


dougmet:
Write out data and metadata to SDMX format.

This will require more information from users. Most notably a DSD (Data Structure Definition) which maps column names and dissagregation categories to UN standards on an indicator by indicator basis.

cf. ONSdigital/sdg-indicators#2891

Start here for specifications https://sdmx.org/?page_id=5008

brockfanning:
Closing this as a duplicate of #52

Issue #9, opened 2019-02-20
Import from SDMX API or remote SDMX files


brockfanning:
Many countries will have their data available in an SDMX API or in remote SDMX files. If we could implement an automated import from these sources, that would help countries get their data into Open SDG (or other) platforms.

So this would be an alternative source of data/metadata, as opposed to the normally expected CSV and YAML. I assume that the output of this import would be the same JSON files that this library already creates?

For those on recent calls about this, I mistakenly thought I had done this before, but I was thinking of the "SDG API", which I tried out [here](https://github.com/GSA/sdg-indicators-usa/pull/861). But this would be something very different and would involve SDMX DSDs.

As a next step we can identify an existing SDMX API (or file) and start thinking about how to convert it to our JSON.

dougmet:
Converting to JSON isn't a problem. Really there's an internal data model and it outputs to whatever. So it's all about reading it into the internal model.

Right now the internal model is really close to the open-sdg output formats. It may be that the internal model needs to change to cope with SDMX and we end up building converters to and from that.

In fact if careful thought is applied this could go together with #4 to make a robust internal data format.

AnnCorp:
From: Daniel Eshetie [mailto:eshetie@un.org] 
Dear Brock,
In case of the Cambodia implementation the draft DSD was used in .Stat suite to publish Cambodia’s SDMX Web service end point as follows
 http://cambodgia-statvm1.eastasia.cloudapp.azure.com/SeptemberDisseminateNSIService 

To get the Global SDGs dataset for a selected indicators for Cambodia, use the following query
http://cambodgia-statvm1.eastasia.cloudapp.azure.com/SeptemberDisseminateNSIService/rest/data/UNSD,DF_SDG,1.0/ 
 
To get their localized datasets including subnational areas based on a customized Global DSD, use the following query:
 http://cambodgia-statvm1.eastasia.cloudapp.azure.com/SeptemberDisseminateNSIService/rest/data/KH_NIS,DF_SDG_KH,1.0/
 
By default, the query returns data in the SDMX-ML 2.1 Generic format. You need to set up the Accept HTTP header to application/json to get the data back as SDMX-json using postman as you see below. Abdulla will add any additional point I missed in case.
 
Regards,
Daniel

brockfanning:
A note - we had discussed focusing first on abstracting the internal model, like @dougmet describes above. I've got an attempt at that here: #16 

[...]

PR #16, opened 2019-03-18
Input/output classes

brockfanning:
Here's a beginning to a possible approach for #4, with a few types of classes:
* 2 "Input" classes: CSV data, and YAML/Markdown metadata
* 1 "Output" class: Open SDG format
* A basic class for indicators

This could use more testing, but a brief run locally seemed to work. Here is a snippet of how it would be used:
```
import os
import sdg

data_pattern = os.path.join('data', '*-*.csv')
data_input = sdg.inputs.InputCsvData(path_pattern=data_pattern)

meta_pattern = os.path.join('meta', '*-*.md')
meta_input = sdg.inputs.InputYamlMdMeta(path_pattern=meta_pattern)

inputs = [data_input, meta_input]

opensdg_output = sdg.outputs.OutputOpenSdg(inputs, output_folder='_site', schema_path='_prose.yml')
outputs = [opensdg_output]

sdg.build_data(outputs=outputs)
```

In theory this is not a breaking change, since it just adds a new optional param to build_data().

I say this is a "beginning" because right now the classes are still using the procedural functions from data.py, json.py, etc.. I assume that we'd probably want all that logic to live inside the appropriate class.

Also, I'm new to Python OOP standards to any feedback or advice is quite welcome.

### DIFF
--- a/sdg/Indicator.py
+++ b/sdg/Indicator.py
@@ -1,38 +1,40 @@
+import sdg
+
 class Indicator:
     """Data model for SDG indicators."""
 
-    def __init__(self, inid, data=None, meta=None, headline=None, edges=None):
+    def __init__(self, inid, data=None, meta=None):
         """Constructor for the SDG indicator instances.
 
         Keyword arguments:
         inid -- The three-part dash-delimited ID (eg, 1-1-1).
         data -- Dataframe of all data.
         meta -- Dict of fielded metadata.
-        headline -- Dataframe of headline data.
-        edges -- Dataframe of data edges.
         """
         self.inid = inid
         self.data = data
         self.meta = meta
-        self.headline = headline
-        self.edges = edges
+        self.set_headline()
+        self.set_edges()
+
+    def has_data(self):
+        """Check to see if this indicator has data."""
+        return False if self.data is None else True
 
     def set_data(self, val):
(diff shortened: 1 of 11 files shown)


dougmet:
Hi Brock, thanks for this. I'd like to spend a little bit of time staring at it before merging in (but I like the approach). I hadn't considered different classes for different types of input (I was essentially thinking of one indicator class with different types of input method). But actually I quite like it, and it should be easy to test too.

brockfanning:
Cool, and to give you some idea of where I'm thinking it might go - other input classes might include InputSdmxApi and InputSdmxFile (#9), and other output classes might include OutputSdmx (#15) and OutputGeoJson (https://github.com/open-sdg/open-sdg/issues/74).

I'll fix the conflicts shortly and update the PR.

brockfanning:
Another note - I didn't carry over the `git_data_dir` parameter into the OOP approach, which is what the US is using to support "wide" CSV data. It would be easy to carry it over, but my thinking was that instead the US might want to first investigate making an InputCsvDataWide subclass.

brockfanning:
Great point, I will refactor that.

brockfanning:
@dougmet This is ready again for you to review.

PR #23, opened 2019-04-01
Release 0.3.1

dougmet:
Patch release to ignore SDMX columns. Only required for users wishing to use SDMX named columns in their data. Currently ignores:

* `Observation status`
* `Unit multiplier`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Might become "release notes"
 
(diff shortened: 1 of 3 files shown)


dougmet:
Hi @brockfanning this is just a quick patch release to cover the SDMX column changes. If it's required for open-sdg 0.5.0 I thought, after this is merged, we should give it a release tag.

PR #27, opened 2019-04-08
Exclude 'Unit measure' column

Norric1Admin:


dougmet:
I'm now wondering if this should be a default argument to a function and then you can pass this in as an option at build time.

Although I don't want to risk this becoming too permanent.

PR #28, opened 2019-04-10
round percentages to 1dp #26

### LINKED ISSUE #26: Stats percentages can add up to more than 100%
In some situations, the percentages for each status in the stats data adds up to more than 100% (probably always 101%, due to rounding?). I'm not sure what the trigger for the bug is, but [here](https://G205SDGs.github.io/sdg-data/stats/reporting.json) is a current example.

dougmet:
A very simple fix. Should fix 99.9% of cases 😉

If we want it more complicated then I suggest we create a `pretty_round` function that does its best to simulate how we'd do it by hand.

Fixes #26

### DIFF
--- a/sdg/stats.py
+++ b/sdg/stats.py
@@ -51,7 +51,7 @@ def reporting_status(schema, all_meta):
         return {'status': status,
                 'translation_key': value_translation[status],
                 'count': count,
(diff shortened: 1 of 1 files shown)


dougmet:
occasionally we might add up to 100.1% or 99.9% but hopefully that's acceptable.

Issue #32, opened 2019-04-28
Usage examples


brockfanning:
Let's start a docs section and add a usage example.

brockfanning:
Closing since an initial example has been merged.

Issue #17, opened 2019-03-20
Schema base class


dougmet:
As discussed in #14 and related to #4 we could extend the `Schema` class so that it has a base class, and then different specific classes for different types of schema input. Each child class would basically override the load function.

brockfanning:
We've got a approach for this in place now, so I will close.

PR #35, opened 2019-05-01
Schema approach using base classes and JSON Schema

brockfanning:
This adds a base class, and a subclass for schemas. It also updates the OutputOpenSdg class to expect one of these classes in the constructor.


brockfanning:
Not ready yet for review.

brockfanning:
Ok, I believe this is ready for review now. I've also added a documentation revamp to this PR, so that it is clear how it would be used.

Fixes #32 and #17.

brockfanning:
Is it odd to throw the defaults directly into the base class? I could separate them out if desired, but it seemed like it is easier inline.

brockfanning:
Actually, hold off on reviewing this -- I am wondering about the possibility of schema for data (as opposed to metadata). I may need to think more about CSV-W so I can better wrap my head around the purpose of schema to these builds.

dougmet:
Apologies for blackout. Will start looking through this shortly. I think csv-w will support column-level metadata, which will be useful for sdmx, and the front end too. But it could just be one supported input/output type.

brockfanning:
@dougmet No prob, and actually hold off for now. I'll ping you once I'm happy with it.

brockfanning:
@dougmet Ok, I think this is ready for you to take a look. I went a lot of different directions, so here are some notes for background:
1. I included a general documentation revamp (well, just a README update and one example script) since it is certainly not obvious how all these classes are intended to work together.
2. Since the input/output class-based builds are sort of a work in progress, and optional, I decided that it would be OK to break backwards-compatibility to a certain extent. So I opted not to worry about translation_key defaults for the schema. Folks will just need to have translation_key set in _prose.yml before trying this.
3. The biggest departure here is the adoption of JSON Schema for the internal data structure of the schema. The upside of this approach is that it's easier to validate metadata, using the 'jsonschema' Python library. It can even handle some of the more “conditional” validations (like what is in check_metadata.py). I've added a validate() method to the Output classes so they can easily validate before running the build.
4. I only worried about validating metadata for now, since JSON Schema will not be much help in validating data.
5. A quick way to test the PR might be to checkout the branch locally, copy docs/examples/open_sdg.py to the root of the project, and then run it (`python open_sdg.py`). This way the "import sdg" will use the code in the current branch.

Norric1Admin:
I've read the files but this is way above my head. @SavvasStephanides any thoughts?

PR #37, opened 2019-05-15
SDMX-JSON and SDMX-ML input

brockfanning:
This is a second draft of the feature. I need to do a bit more testing before merging this.


brockfanning:
@dougmet The Python 3.4 tests were failing with this, I'm guessing because of the "requests" library. So this PR also drops Python 3.4 from the test suite. Do you think this is a problem? Should I try something besides "requests" to see if we can keep 3.4 support? Or is it OK to go ahead and require 3.5 or higher?

brockfanning:
@AnnCorp @Norric1Admin @calebdow @kmrebello @philipashlock Would appreciate if you all could take a look at this [Cambodia proof-of-concept](http://brock.tips/sdg-site-cambodia/). A few notes about it:
1. The [data repository](https://github.com/brockfanning/sdg-data-cambodia) has no data/metadata files - everything is being pulled in from an SDMX JSON API. It's only pulling in data - no metadata at the moment.
2. The [site repository](https://github.com/brockfanning/sdg-site-cambodia) has no `_indicators` or `_goals` folders - the indicators/goals are being created based on the indicators present in the source data.
3. Because the source data includes indicators from a goal 18, the goals go all the way to 18. But the "goal 18" icon is still the same "Sustainable Development Goals" logo.

Would appreciate any feedback on the data/graphs. I'm hoping this will help identify any shortcomings in the platform and/or problems in this SDMX import code.

dougmet:
I've been looking for a good excuse to drop 3.4 for a when now. I'm sure there's a beat around requests but one way or another it's causing hassle. It comes with an old pandas version too which changes its API a lot so also breaks.

In short, I don't see any problem dropping 3.4 support. We'll need to update it in the setup.py etc to enforce this.

kmrebello:
Hi @brockfanning. I noticed that the line graphs for 2.2.1, 7.1.1, and 12.4.1 are coming up blank. In the csv files, I'm seeing the year column but nothing else is visible.

brockfanning:
@kmrebello That does seem to be a bug in Open SDG. When an indicator has no "headline" (ie, no aggregate series of country-wide stats) and includes more than 1 disaggregation, the graph starts off blank. For example, as you saw, [2.2.1](http://brock.tips/sdg-site-cambodia/2-2-1/) shows a blank graph, until a "Reference area" is selected.

You can see from the [source data](http://brock.tips/sdg-data-cambodia/data/2-2-1.csv) that there are no rows with less than 2 disaggregations. I think if we were manually maintaining this, we'd probably remove the "Age" column altogether. And maybe this SDMX importer should automatically drop disaggregations that have only 1 variation.

Regardless I think Open SDG should be able to handle cases like this, so I've created an issue for it: https://github.com/open-sdg/open-sdg/issues/174

brockfanning:
@kmrebello Also about 7.1.1 - that I think is another bug in Open SDG, which I've started an issue for [here](https://github.com/open-sdg/open-sdg/issues/175).

brockfanning:
That's a good point @dougmet. I moved a couple of small chunks of code into individual methods - and also moved them into the base class, since they are probably useful to any type of input.

But the execute function is still quite hard to read. The challenge is that in the SDMX-JSON format, pretty much every value is actually an integer referring to the position of some item in another array, elsewhere in the JSON. But I bet there is a way to improve this, so I'll keep thinking on it.

Norric1Admin:
A few thoughts on this (you may be aware of them all)
- No headline data ([9.1.2.1](http://brock.tips/sdg-site-cambodia/9-1-2-1/), [3.4.1](http://brock.tips/sdg-site-cambodia/9-1-2-1/), etc.): 'Cambodia' has been listed a disaggregation under reference area, but should ideally be a blank in the CSV.
- [3.8.1](http://brock.tips/sdg-site-cambodia/3-8-1/) showing no data unless all 3 categories checked.
- Minor point but 'Percent' should ideally have the symbol after it in brackets ([4.2.1](http://brock.tips/sdg-site-cambodia/4-2-1/)).

brockfanning:
@Norric1Admin Could you elaborate about the general weirdness? I'd like to know if anything in this import may be causing an issue.

brockfanning:
A couple things I am wondering about:
* Is the [id mapping](https://github.com/brockfanning/sdg-data-cambodia/blob/develop/scripts/output.py#L9) maybe available in the DSD?
* I believe translations are being added - are they also in the DSD, and if so, can/should we import them here?

brockfanning:
@dougmet The latest commit should not change any functionality (hopefully) but refactors to make it clearer what's going on. It also helped me notice some inefficiencies.

Norric1Admin:
> @Norric1Admin Could you elaborate about the general weirdness? I'd like to know if anything in this import may be causing an issue.

@brockfanning Mainly that the reporting status categories are showing as the metadata values (`complete`, `notstarted`) instead of their aliases (e.g. 'Exploring Data Sources'). The green `complete` bar should ideally be first (green, yellow, red) and the bonus Goal 18 has 6 indicators (should be hidden).

brockfanning:
Thanks @Norric1Admin. I've fixed the problems with the categories. The issue with Goal 18 is actually expected, as the data source includes a Goal 18 with 6 indicators in it.

brockfanning:
Just a note that I've got feedback from the POC which implies that the refactor did not break anything. :)

One thing I'd like to tweak is the indicator ID mapping. Maybe this library should have a hardcoded list of all the global ID mappings, so that countries only have to provide their national ID mapping.

Norric1Admin:
@brockfanning How is it looking now you've added these changes? https://github.com/open-sdg/sdg-build/pull/37/commits/8b20d4b618b641a395d6f98d3fccf5e1fff0b771

brockfanning:
@Norric1Admin It now incorporates the DSD to get the list of indicators to import. I'm now working on the SDMX-ML import. I think that should probably be done at the same time. In fact I'll update the name of this issue accordingly.

brockfanning:
@LucyGwilliamAdmin I believe this is now working for both SDMX-JSON and SDMX-ML "Structure". I was thinking we could also add support for "Structure Specific" in this PR, which should not be too hard.

I do have one question - looking at [this DSD](https://raw.githubusercontent.com/LucyGwilliamAdmin/SDMX/master/SDG_DSDv0.7.xml) I see that the codelist ID for the series is "CL_SDG_SERIES", but looking at the latest [UN DSD](https://unstats.un.org/sdgs/files/SDG_DSD.xml) it is "CL_SERIES". Do you know if we need to make this configurable? Or is one or the other of those DSDs out of date?

LucyGwilliamAdmin:
@brockfanning great news! What would 'Structure Specific' involve?

Yes, I would go with CL_SERIES as that's what is used in DSD V1.0

LucyGwilliamAdmin:
Does this mean a SDMX-ML file could be uploaded to a repo once this is merged? Is there somewhere I can test?

brockfanning:
@LucyGwilliamAdmin Yes this is ready for testing. I think, given this object-oriented approach is still largely unused, I'll go ahead and merge this. It's at a decent stopping place, with support for SDMX-JSON, SDMX-ML Structure, and SDMX-ML Structure Specific. There may still be opportunities for refactoring and improvements/fixes.

This includes some updates to the README and demos in the docs/examples folder.

LucyGwilliamAdmin:
@brockfanning can you explain how this would work with SDMX-ML? I want to try and upload an SDMX-ML file directly

brockfanning:
@LucyGwilliamAdmin Sure thing. See [this example](https://github.com/open-sdg/sdg-build/blob/master/docs/examples/sdmx_ml.py). The `source` parameter there is a remote file, but it could also be a path to a file on disk, eg:

```
source = 'my-folder/my-file.xml'
```

Though to be platform-agnostic you might do:

```
source = os.path.join('my-folder', 'my-file.xml')
```

The same goes for the `dsd` parameter.

You'll want to read through the comments and tweak/remove/add other parameters as needed.

Note that the file you end up with is meant to be a replacement for the `build_data.py` files that the data repos have. ([example](https://github.com/ONSdigital/sdg-data/blob/develop/scripts/build_data.py))

brockfanning:
@LucyGwilliamAdmin Also for a real-world example, [here](https://github.com/brockfanning/sdg-data-cambodia/tree/develop/scripts) is a data repo that's using this type of approach (other than the fact that its SDMX source and DSD are remote instead of local).

LucyGwilliamAdmin:
@brockfanning is there a way for this to be set up in a similar way to uploading CSVs e.g. can upload SDMX-ML files to the data folder with a naming convention (e.g. "1.1.1.xml")?

brockfanning:
@LucyGwilliamAdmin I haven't tested this, but in theory something like this would work:
```
import os
import sdg

# Control how the SDMX dimensions are mapped to Open SDG output. Because the
# Open SDG platform relies on a particular "Units" column, we control that here.
dimension_map = {
    # Open SDG needs the unit column to be named specifically "Units".
    'UNIT_MEASURE': 'Units',
}

# Each SDMX source should have a DSD (data structure definition).
dsd = os.path.join('path', 'to', 'my', dsd')

# The "indicator_id" (such as 1-1-1, 1-2-1, etc.) is not yet formalized in the
# SDG DSD standard. It is typically there, but it's location is not predictable.
# So, specify here the XPath query needed to find the indicator id inside each
# series code. This is used to map series codes to indicator ids.
indicator_id_xpath = ".//CHANGEME"
indicator_name_xpath = ".//CHANGEME"

# Read all the files.
sdmx_files = glob.glob(os.path.join('path', 'to', 'sdmx', 'files', '*.xml'))
inputs = []
for sdmx_file in sdmx_files:
    # Create the input object.
    data_input = sdg.inputs.InputSdmxMl_Structure(
        source=sdmx_file,
        dimension_map=dimension_map,
        dsd=dsd,
        indicator_id_xpath=indicator_id_xpath,
        indicator_name_xpath=indicator_name_xpath
    )
    inputs.append(data_input)

# Use the Prose.io file for the metadata schema.
schema_path = os.path.join('tests', '_prose.yml')
schema = sdg.schemas.SchemaInputOpenSdg(schema_path=schema_path)

# Create an "output" from these inputs and schema, for JSON for Open SDG.
opensdg_output = sdg.outputs.OutputOpenSdg(inputs, schema, output_folder='_site')

# Validate the indicators.
validation_successful = opensdg_output.validate()

# If everything was valid, perform the build.
if validation_successful:
    opensdg_output.execute()
```

The main difference is that is there is a loop for the inputs, instead of a single input. It's a little inefficient in that it reloads the DSD for every indicator, but should work.

Note that the `os.path.join`, `indicator_id_xpath`, and `indicator_name_xpath` would need to be changed as appropriate.

LucyGwilliamAdmin:
@brockfanning I have just tried this [here](https://github.com/ONSdigital/sdg-data/tree/sdmx) but I'm getting this error:
image

brockfanning:
@LucyGwilliamAdmin Looks like it's not able to find the "Series" where it expects. Is it possible you're using a "structure specific" SDMX format? If so, you'd use `InputSdmxMl_StructureSpecific` instead of `InputSdmxMl_Structure`.

LucyGwilliamAdmin:
@brockfanning just checked and the files are structure specific so changed that, but getting this now:
image

brockfanning:
@LucyGwilliamAdmin It looks like it's interpreting the goal of that indicator as "Proportion", when it should be "1". This actual cause is that it can't figure out what the indicator id (1.1.1 in this case) should be. It's incorrectly interpreting "Proportion" as the indicator id, and so the goal (which is interprets from the indicator id) is also wrong.

For the proof-of-concept I did (the SDMX from Cambodia) the indicator ID was included in the `Name` (for example: `4.3.1 Gross enrolment rate in tertiary education (aged 18-21 years) by sex`). But it looks like in your SDMX the indicator IDs are not in the DSD anywhere.

Though it seems like we should be able to grab the indicator ID from the file name, I think in many cases SDMX sources will not be split up by indicator. But we could re-think that if I'm off base there.

A quick fix would be to hardcode the indicator ids, using the `indicator_id_map` parameter. You can see an example of using it [here](https://github.com/brockfanning/sdg-data-cambodia/blob/develop/scripts/output.py#L23) and [here](https://github.com/brockfanning/sdg-data-cambodia/blob/develop/scripts/output.py#L32). In your case the parameter would consist of:
```
indicator_id_map = {
    'SI_POV_DAY1': '1-1-1'
}
```

LucyGwilliamAdmin:
Ok, great thanks - got it working [here](https://sdg.mango-solutions.com/sdmx-files/1-1-1/)
So that dictionary should consist of all indicators so that Kyrgyzstan can upload all of their data in SDMX format?

brockfanning:
@LucyGwilliamAdmin Unless there is some element in the DSD that includes the indicator IDs, yes, that `indicator_id_map` is the only way to tell sdg-build which series belong to which indicator.

LucyGwilliamAdmin:
@brockfanning what about indicators that don't have data? Similarly to having CSV as the input, would we need to SDMX-ML placeholders?

Issue #55, opened 2019-10-14
Possible problem with edges.py


brockfanning:
I'm not sure if this is an invalid case, or just an edge-case that hasn't happened yet. Or maybe I'm misinterpreting something. In my case, here is the [CSV](https://github.com/brockfanning/sdg-data-cambodia/blob/gh-pages/data/1-2-1.csv) and here are the [edges](https://github.com/brockfanning/sdg-data-cambodia/blob/gh-pages/edges/1-2-1.csv). Shouldn't the edges be empty? Why is it generating a "Reference area" to "Degree of urbanisation" edge?

brockfanning:
Looking at the code, actually I think [this condition](https://github.com/open-sdg/sdg-build/blob/master/sdg/edges.py#L12) is being matched by the CSV above: "Reference area" is sometimes present when "Degree of urbanisation" is not present. So edges.py is working as intended. I want to figure out if that condition could be made more specific to avoid this problem, though.

brockfanning:
Not technically a problem with edges.py. But there is code in edges.py that expects empty values to be np.NaN, and I was instead setting them to empty strings.

PR #56, opened 2019-10-14
Convert empty strings or None to np.NaN

brockfanning:
This avoids a problem when a user replaces data with an empty string, expecting that it will make the cell empty. The problem manifests in edges.py in the x_without_y function, where an empty string is not treated as null.


brockfanning:
Fixes #55

PR #63, opened 2019-11-07
Translated builds

brockfanning:
This adds an optional way to generate fully-translated builds. If the `execute()` method is called with a language parameter then the build will be translated into that language and placed into a subfolder. For example, `execute('en')` will translate everything (metadata and data) into English and place the full build into an `en` subfolder.

This also includes a helper method for doing multiple languages at once: `execute_per_language`, which should get a list of languages. Eg: `execute_per_language(['ru', 'es'])`

If the `execute` method is used without any parameters (as normal), then the build is not translated, and is placed in the root folder. So, the way things work now will continue to work, which means that this PR has no breaking changes. (only additional functionality)


brockfanning:
@LucyGwilliamAdmin Would you be able to try this branch and see if it fixes the problem with reporting status not getting translated? (https://github.com/open-sdg/open-sdg/issues/389)

PR #64, opened 2019-11-13
Add ZIP export functionality

SavvasStephanides:
This change:

* Creates a `IndicatorExportService` which is responsible for exporting all indicator CSV files as a ZIP file 
* Amends the `build_data` function to include the ZIP export logic.

Test link: 

https://sdg.mango-solutions.com/data/ziptest/

In this feature branch a `zip` folder has been created which contains a file `all_indicators.zip` containing all indicator CSV files.


SavvasStephanides:
> You've called it in the "procedural" approach to builds, but can you also call it in the object-oriented approach. That would be basically around this line.

Will do

> Shouldn't it be added to this file so that it can be included in a general import sdg?

Does this mean just moving [this line](https://github.com/open-sdg/sdg-build/blob/29e196f1b5db32d93a8389e842fdaa0d1930d1fe/sdg/build.py#L15) to [this file](https://github.com/open-sdg/sdg-build/blob/master/sdg/__init__.py)?

brockfanning:
> > Shouldn't it be added to this file so that it can be included in a general import sdg?
> 
> Does this mean just moving [this line](https://github.com/open-sdg/sdg-build/blob/29e196f1b5db32d93a8389e842fdaa0d1930d1fe/sdg/build.py#L15) to [this file](https://github.com/open-sdg/sdg-build/blob/master/sdg/__init__.py)?

Basically yes, but to follow the existing examples I think it would be:
```
from .IndicatorExportService import IndicatorExportService
```

SavvasStephanides:
Thanks Brock! Updated with requested changes.

PR #65, opened 2019-11-16
Allow configuration of subfolder in translation input

brockfanning:
This makes the translation input more flexible by allowing it to be used with differing subfolders.


brockfanning:
On second thought, this doesn't a whole lot of flexibility. I'll close for now unless this becomes needed for a particular use-case.

PR #77, opened 2019-11-25
1.0.0-rc1 - default to using the object-oriented API

brockfanning:
This PR started out as an attempt at switching to the object-oriented API by default, but has grown to include a more thorough preparation for version 1.0.0. Here is a summary of what is included in this PR:

1. When the legacy Open SDG helper functions are used, now the object-oriented API will be used. These legacy helper functions are `build_data`, `check_all_csv`, and `check_all_meta`. They should continue to function as before (with one exception - see breaking changes below) but will be using the object-oriented approach behind the scenes.
2. This can be seen as a big step towards a 1.0.0 release, so this also bumps the version of the library to 1.0.0-rc1.
3. Open SDG configuration files: This feature adds the capability of managing the configuration of an Open SDG implementation via a config file. This feature is intended to make it easier for countries using the out-of-the-box CSV/YAML approach to tweak their configuration, without requiring them to understand Python code. For an example of the options, see `docs/examples/open_sdg_config.yml`. For an example of using the config file, see `docs/examples/open_sdg_simple.py`.
4. This PR also includes a refresh of the testing metadata - pulling in the latest metadata from open-sdg-data-starter.
5. Some classes, scripts, and tests that are deprecated by the object-oriented approach have been removed, for clarity.
6. Update the default translation repo from sdg-translations to multiple: translations-un-sdg and translations-open-sdg.

## How to test

The best way to test would be to simply point to this branch (`brockfanning/sdg-build-1@default-to-new-api`) in an Open SDG data repository's requirements.txt file. If all goes well, it should work as a "drop in replacement" (but see the breaking changes below).

## Breaking changes

1. The `build_data` helper function no longer supports the `git` or `git_data_dir` parameters. The new behavior is to run the function as if `git` were True. For most implementations this should not present a problem, but please open an issue in GitHub if you encounter any unintended consequences.
2. Because this switches to a new (slightly more strict) validation system, you may find that your metadata fails validation after updating. However the failure messages should include details on which indicators failed and why.


brockfanning:
@LucyGwilliamAdmin @dougmet Curious what you think about this one. The idea here is to update the `build_data` and `check_all_meta` functions so that they use the OOP approach, behind the scenes. So in theory, countries could start using the OOP approach without making any changes.

One thing to mention is that this drops support for the `git` and `git_data_dir` parameters (see [here](https://github.com/open-sdg/sdg-build/blob/master/sdg/build.py#L24)) because they were not implemented in the OOP approach. The `git` parameter is just assumed to be True, and the `git_data_dir` functionality is gone altogether.

EDIT: I'm realizing that there is nothing in the OOP approach yet for validating data (as opposed to metadata). I'll create an issue for that, and this should probably wait for that.

brockfanning:
Another thing this ought to include is support for a YAML config file. That would make it easier for countries to use some of the options in the OOP approach - like translated builds.

brockfanning:
Hold off on reviewing this - I'm going to add to it today.

brockfanning:
@LucyGwilliamAdmin @SavvasStephanides I've been adding little tweaks/refactors but I believe this is complete and ready for review. This PR is intended to get sdg-build to a 1.0.0-rc1 release. See details in the PR description, including breaking changes and instructions on how to test.

SavvasStephanides:
I like the idea to use OOP methods by default. If I understand correctly, are the old functions now merely a call to the OOP functions for backwards compatibility?

Also can you resolve the conflicts in `SchemaInputBase.py`?

brockfanning:
@SavvasStephanides Yep, backwards compatibility and also just as helper functions specifically for Open SDG.

That conflict should be resolved now.

brockfanning:
I think it's time to "pull the trigger" on this one. I'll push a 1.0.0-rc1 tag after merging.

PR #80, opened 2020-01-04
Allow reporting stats to include extra fields

brockfanning:
This is needed for [this issue](https://github.com/open-sdg/open-sdg/issues/209). It is not a breaking change - it only adds an "extra_fields" property in the reporting stats.

## How it works

An Open SDG site's "data repository" needs to be configured to add "extra_fields" to the reporting stats output. This can be done in the OutputOpenSdg class with a `reporting_status_extra_fields` parameter. For example, instead of this:
```
opensdg_output = sdg.outputs.OutputOpenSdg(
    inputs=inputs,
    schema=schema,
    output_folder='_site',
    translations=translations)
```
You would do this:
```
opensdg_output = sdg.outputs.OutputOpenSdg(
    inputs=inputs,
    schema=schema,
    output_folder='_site',
    translations=translations,
    reporting_status_extra_fields=['an_extra_field', 'another_extra_field'])
```

### DIFF
--- a/docs/examples/open_sdg_config.yml
+++ b/docs/examples/open_sdg_config.yml
@@ -35,3 +35,9 @@ site_dir: _site
 # This identifies a file containing the schema (possible fields) for metadata.
 # Currently this needs to be a prose.io config, and defaults to '_prose.yml'.
 schema_file: _prose.yml
+
+# Reporting status extra fields
+# -----------------------------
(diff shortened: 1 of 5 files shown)


brockfanning:
This will also fail testing until the tests are updated. But I'll hold off to avoid conflicts with #77

PR #91, opened 2020-01-23
Allow SDMX to import translation keys instead of text

### LINKED ISSUE #90: SDMX translations to multilingual Open SDG
Although we've already implemented the import of translations from an SDMX DSD, we haven't yet figured out how to integrate those translations with Open SDG output. The obvious approach would be to use the SDMX ids as the "translation keys", and then import SDMX ids instead of human-readable text during the import. One catch may be if the SDMX ids are not unique enough. For example, if the dimension value ids are not unique enough, they would have to each be in their own translation group.

brockfanning:
Fixes #90 

This adds an `import_translation_keys` parameter to the SDMX inputs, which instructs them to pull in "translation keys" rather than fully-translated text. These follow the format of "`code.`" plus the id of the code, and "`concept.`" plus the id of the concept. This convention matches what is generated by this library's TranslationInputSdmx class, allowing the two to be used in combination.

For example, rather than pulling in "Per 1,000 live births", it would pull in "code.PER_1000_LIVE_BIRTHS" instead. At the same time, the TranslationInputSdmx class would have pulled in a translation structure like this:
```
{
    "en": {
        "code": {
            "PER_1000_LIVE_BIRTHS": "Per 1,000 live births"
        }
    }
}
```

### DIFF
--- a/docs/examples/sdmx_json.py
+++ b/docs/examples/sdmx_json.py
@@ -35,6 +35,7 @@ data_input = sdg.inputs.InputSdmxJson(
     source=source,
     dimension_map=dimension_map,
     dsd=dsd,
+    import_translation_keys=True,
     indicator_id_map=indicator_id_map,
     indicator_id_xpath=indicator_id_xpath,
     indicator_name_xpath=indicator_name_xpath
--- a/docs/examples/sdmx_json.py
+++ b/docs/examples/sdmx_json.py
@@ -50,8 +50,8 @@ translations = [
     # Use two Git repos containing translations.
     sdg.translations.TranslationInputSdgTranslations(source='https://github.com/open-sdg/translations-un-sdg.git', tag='1.0.0-rc1'),
     sdg.translations.TranslationInputSdgTranslations(source='https://github.com/open-sdg/translations-open-sdg.git', tag='1.0.0-rc2'),
-    # Also look for translations in a local 'translations' folder.
-    sdg.translations.TranslationInputYaml(source='translations'),
+    # Also pull in translations from the SDMX DSD.
+    sdg.translations.TranslationInputSdmx(source=dsd),
 ]
 
 # Create an "output" from these inputs and schema, for JSON for Open SDG.
--- a/docs/examples/sdmx_ml.py
+++ b/docs/examples/sdmx_ml.py
@@ -34,6 +34,7 @@ data_input = sdg.inputs.InputSdmxMl_Structure(
     dimension_map=dimension_map,
     dsd=dsd,
     drop_dimensions=drop_dimensions,
+    import_translation_keys=True,
     indicator_id_xpath=indicator_id_xpath,
     indicator_name_xpath=indicator_name_xpath
 )
--- a/docs/examples/sdmx_ml.py
+++ b/docs/examples/sdmx_ml.py
@@ -48,8 +48,8 @@ translations = [
     # Use two Git repos containing translations.
     sdg.translations.TranslationInputSdgTranslations(source='https://github.com/open-sdg/translations-un-sdg.git', tag='1.0.0-rc1'),
     sdg.translations.TranslationInputSdgTranslations(source='https://github.com/open-sdg/translations-open-sdg.git', tag='1.0.0-rc2'),
-    # Also look for translations in a local 'translations' folder.
-    sdg.translations.TranslationInputYaml(source='translations'),
+    # Also pull in translations from the SDMX DSD.
+    sdg.translations.TranslationInputSdmx(source=dsd),
 ]
 
 # Create an "output" from these inputs and schema, for JSON for Open SDG.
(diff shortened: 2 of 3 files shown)


brockfanning:
@LucyGwilliamAdmin To test this, you could want to do a few things:
1. Add an instance of TranslationInputSdmx to the data build, like in the example scripts in this PR.
2. I would recommend doing translated by builds, by using `execute_per_language` instead of just `execute`. For an example see the files changed in #93.
3. Add `import_translation_keys=True` in the creation of the Output, like in the example scripts in this PR.

LucyGwilliamAdmin:
@brockfanning I've started taking a look at this on the site I created to test the new quick start guidance but with Kyrgyzstan's data, metadata, etc.

Even though the data is being built at ru and en e.g. https://lucygwilliamadmin.github.io/sdg-data-quick-start/ru/meta/1-1-1.json, the changes aren't coming through to the site: https://lucygwilliamadmin.github.io/sdg-site-quick-start/reporting-status/ (even after forcing a rebuild of the site)

brockfanning:
@LucyGwilliamAdmin Everything looked good as far as your config goes. I just submitted a PR that might help.

brockfanning:
@LucyGwilliamAdmin Another problem you'll run into, it appears that the DSD you're using is not fully translated into Russian. For example, here is the concept element for REF_AREA:

```
        <structure:Concept id="REF_AREA" urn="urn:sdmx:org.sdmx.infomodel.conceptscheme.Concept=KG_NSC:SDG_CONCEPTS(1.0).REF_AREA">
          <common:Name xml:lang="en">Reference area</common:Name>
          <structure:CoreRepresentation>
            <structure:Enumeration>
              <Ref id="CL_AREA" version="1.0" agencyID="KG_NSC" package="codelist" class="Codelist" />
            </structure:Enumeration>
          </structure:CoreRepresentation>
        </structure:Concept>
```

Notice it only has an English name, but not a Russian name.

The effect this has is that, while the English build is getting translated into English, the Russian build is left with just translation keys. For example, compare:
* [English data for 1-1-1](https://github.com/LucyGwilliamAdmin/sdg-data-quick-start/blob/gh-pages/en/data/1-1-1.csv)
* [Russian data for 1-1-1](https://github.com/LucyGwilliamAdmin/sdg-data-quick-start/blob/gh-pages/ru/data/1-1-1.csv)

LucyGwilliamAdmin:
@brockfanning I've just merged that PR and the metadata is coming through fine now (sort of - metadata field names in metadata section aren't showing properly). The data isn't coming through - getting this error in the console for [1.1.1](https://lucygwilliamadmin.github.io/sdg-site-quick-start/ru/1-1-1/) (although could that be something to do with translations not being available as you mentioned above?):
image

I will email Abdulla to see if he can send latest Kyrgyzstan DSD with translations

brockfanning:
@LucyGwilliamAdmin I'm not seeing that error in my console when I look at the page. Could you maybe try a hard-refresh (shift-F5) and confirm that you consistently see that error? If so it could be a bug, or maybe the site repo needs a rebuild.

LucyGwilliamAdmin:
@brockfanning I've just done a hard refresh and data is showing now - I was too impatient!

However, no metadata is showing at all.

brockfanning:
That's interesting about the metadata. Nothing in this PR should have affected metadata, but maybe it was already a bug with the translated build approach. I'll look into that.

brockfanning:
@LucyGwilliamAdmin In the site repo, try updating to the latest version of jekyll-open-sdg-plugins (1.0.0.rc15). I think this should fix the metadata problem.

LucyGwilliamAdmin:
@brockfanning great - that worked. Think we need DSD with translations before we can confirm all ok.

I do have one question though - I know there's a function somewhere that makes sure that Kyrgyzstan data appears as headline data. However on my test site - this is not the case e.g. [1.1.1](https://lucygwilliamadmin.github.io/sdg-site-quick-start/ru/1-1-1/). To overcome this do I need to change the function so that code.KG is always headline?

brockfanning:
That's a good question.

How would you feel about a solution where "code.KG" was simply changed to "" during the data build?

brockfanning:
@LucyGwilliamAdmin If you want to try that out, I believe the way you would do it is to add the following to the "dimension_map":

```
'REF_AREA|KG': ''
```

So you'd add that below this line: https://github.com/LucyGwilliamAdmin/sdg-data-quick-start/blob/develop/scripts/output.py#L39

LucyGwilliamAdmin:
@brockfanning Yes that worked. Nazira has just sent over DSD so will try that now

LucyGwilliamAdmin:
@brockfanning have just uploaded latest DSD - translations only seem to be working for breakdown names and not disagreggation names (see [1.2.1](https://lucygwilliamadmin.github.io/sdg-site-quick-start/ru/1-2-1/))

brockfanning:
@LucyGwilliamAdmin I think that means that the "Concepts" are not translated in the DSD.

LucyGwilliamAdmin:
@brockfanning yes I've just checked and that is the case - are you happy that this is working as it should be?

brockfanning:
@LucyGwilliamAdmin Yep, I'm pretty happy with how it's working.

PR #98, opened 2020-02-07
Only zip indicators that are complete and statistical

brockfanning:
Here is a quick attempt at fixing https://github.com/open-sdg/open-sdg/issues/469.

### DIFF
--- a/sdg/Indicator.py
+++ b/sdg/Indicator.py
@@ -238,7 +238,9 @@ class Indicator:
     def is_complete(self):
         """Decide whether this indicator can be considered "complete".
 
-        Returns : boolean
+        Returns
+        -------
+        boolean
             True if the indicator can be considered "complete", False otherwise.
         """
         # First, check for an open-sdg-style "reporting_status" metadata field,
@@ -252,6 +254,27 @@ class Indicator:
         # Otherwise fall back to whether the indicator has data and metadata.
         else:
             return self.has_data() and self.has_meta()
+
+
+    def is_statistical(self):
+        """Decide whether this indicator can be considered "statistical".
+
+        Returns
+        -------
+        boolean
+            True if the indicator can be considered statistical, False otherwise.
+        """
+        # First, check for an open-sdg-style "data_non_statistical" metadata field.
+        non_statistical = self.get_meta_field_value('data_non_statistical')
+        if non_statistical is None or non_statistical == False:
+            return True
+        # If the the indicator was explicitly non-statistical, return False.
+        elif non_statistical == True:
+            return False
+        # Otherwise fall back to whether the indicator has data.
+        else:
+            return self.has_data()
+
 
     def get_meta_field_value(self, field):
         """Get the value for a metadata field.
--- a/sdg/Indicator.py
+++ b/sdg/Indicator.py
@@ -253,19 +253,17 @@ class Indicator:
         else:
             return self.has_data() and self.has_meta()
 
-    def get_meta_field_value(self, field, language=None):
+    def get_meta_field_value(self, field):
         """Get the value for a metadata field.
 
         Parameters
         ----------
         field : string
             The key of the metadata field.
-        language : string or None
-            A language to use for translation, if any.
 
         Return : string or None
-            The value of the specified field in the specified language, or
-            just None if the field could not be found.
+            The value of the specified field or just None if the field could not
+            be found.
         """
 
         if not self.has_meta():
@@ -274,4 +272,4 @@ class Indicator:
         if field not in self.meta:
             return None
 
-        return self.language(language).meta[field]
+        return self.meta[field]
--- a/sdg/Indicator.py
+++ b/sdg/Indicator.py
@@ -234,3 +234,44 @@ class Indicator:
 
         # Finally place the translation for later access.
         self.translations[language] = indicator
+
+    def is_complete(self):
+        """Decide whether this indicator can be considered "complete".
+
+        Returns : boolean
+            True if the indicator can be considered "complete", False otherwise.
+        """
+        # First, check for an open-sdg-style "reporting_status" metadata field,
+        # for a value of "complete".
+        reporting_status = self.get_meta_field_value('reporting_status')
+        if reporting_status is not None and reporting_status == 'complete':
+            return True
+        # If there was some other reporting status, assume not complete.
+        elif reporting_status is not None:
+            return False
+        # Otherwise fall back to whether the indicator has data and metadata.
+        else:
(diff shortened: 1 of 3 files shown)


SavvasStephanides:
Not sure if I've done something wrong during testing, but it seems that it still loads non-complete indicators on the zip file. I've basically updated the last line of `requirements.txt` to:

```
git+git://github.com/brockfanning/sdg-build-1@zip-complete-indicators-only
```

[And here's the end result](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/zip_filter_test/zip/)

brockfanning:
@SavvasStephanides What's an example of an incomplete indicator that's getting included in the zip?

SavvasStephanides:
`1-5-3`.

There are a few files that are just "This is a non-statistical indicator". They probably shouldn't be included in the zip(?)

brockfanning:
@SavvasStephanides I made a change to also exclude the non-statistical indicators. Can you try again?

SavvasStephanides:
@brockfanning Great, it works now! I'll have a peak at the code and approve 👍

sdg/Indicator.py
```diff
@@ -234,3 +234,65 @@ def translate_data(text):
...
         self.translations[language] = indicator
+
+    def is_complete(self):
+        """Decide whether this indicator can be considered "complete".
+
+        Returns
+        -------
+        boolean
+            True if the indicator can be considered "complete", False otherwise.
+        """
```
SavvasStephanides: Any specific reason we need these comment blocks? The purpose of the function and its return type seems to be self-explanatory from the name

sdg/Indicator.py
```diff
@@ -234,3 +234,65 @@ def translate_data(text):
...
+        # First, check for an open-sdg-style "reporting_status" metadata field,
+        # for a value of "complete".
```
SavvasStephanides: Please focus on making the code readable, before considering writing comments. I think for this specific example, the code below is pretty well self-documented.

sdg/IndicatorExportService.py
```diff
@@ -36,6 +47,14 @@ def __get_all_indicator_csv_files(self):
     def __file_is_csv(self, file_name):
         return file_name.endswith(".csv")
 
+    def __file_has_complete_statistical_data(self, file_name):
+        indicator_id = file_name.split('.')[0]
+        if indicator_id not in self.__indicators:
+            raise KeyError("Could not check whether %s is complete." % indicator_id)
+        complete = self.__indicators[indicator_id].is_complete()
+        statistical = self.__indicators[indicator_id].is_statistical()
+        return complete and statistical
```
SavvasStephanides: Good thinking separating this into its own function. Maybe this could be done in `Indicator.py` instead though?

sdg/outputs/OutputOpenSdg.py
```diff
@@ -79,7 +79,7 @@ def build(self, language=None):
         stats_reporting = sdg.stats.reporting_status(self.schema, all_meta, self.reporting_status_grouping_fields)
         status = status & sdg.json.write_json('reporting', stats_reporting, ftype='stats', site_dir=site_dir)
 
-        indicator_export_service = sdg.IndicatorExportService(site_dir)
+        indicator_export_service = sdg.IndicatorExportService(site_dir, self.indicators)
```
SavvasStephanides: Good idea passing the indicators into the service.

sdg/Indicator.py
```diff
@@ -234,3 +234,65 @@ def translate_data(text):
...
+
+
+    def is_statistical(self):
+        """Decide whether this indicator can be considered "statistical".
+
+        Returns
+        -------
+        boolean
+            True if the indicator can be considered statistical, False otherwise.
+        """
```
SavvasStephanides: As above, regarding comment blocks.

sdg/Indicator.py
brockfanning: @SavvasStephanides I wanted to make clear that we are looking for something specific to open-sdg. Remember that sdg-build (in theory at least) is not limited to open-sdg stuff. So, normally I would refrain from putting open-sdg-specific code in Indicator.py. But since it's somewhat harmless and time is short, I decided to go ahead and do that, along with a comment to make clear what's going on.

sdg/IndicatorExportService.py
brockfanning: @SavvasStephanides Do you mean an `is_complete_and_statistical` method in Indicator.py? Personally I would find that redundant, since it already has `is_complete` and `is_statistical`.

sdg/Indicator.py
brockfanning: @SavvasStephanides I try to always do doc blocks regardless (though sometimes I forget). It's not a huge deal, but if we ever do automatically-generated documentation, we'll be glad to have them. And in general it's better to have them than not.

sdg/IndicatorExportService.py
SavvasStephanides: Yes, I mean have it in `Indicator.py` and pass *that* function here.

Maybe something like `is_suitable_for_export`, or something similar (it needs to be public to be then accessed by the service. This would encapsulate checking for suitability of an indicator to a single function, plus it would be potentially easy to extend when we need to add additional checks in the future.

sdg/Indicator.py
SavvasStephanides: Fair enough. Just be consistent and remember to update when appropriate.

brockfanning: @SavvasStephanides I should make clear that I'm far from a Python expert, and would be happy to adopt whatever coding and documentation standards we agree on. And I definitely agree that every code change should always include any necessary documentation changes.

sdg/IndicatorExportService.py
brockfanning: @SavvasStephanides I just updated that method - with the idea of adding to it in the future if we need more limits on what gets exported.

sdg/IndicatorExportService.py
```diff
@@ -32,7 +32,7 @@ def __get_all_indicator_csv_files(self):
         all_data_file_names = os.listdir(self.__data_directory)
         csv_data_file_names = []
         for file_name in all_data_file_names:
-            if self.__file_is_csv(file_name) and self.__file_has_complete_statistical_data(file_name):
+            if self.__file_is_csv(file_name) and self.__file_is_suitable_for_export(file_name):
```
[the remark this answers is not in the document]
brockfanning: @SavvasStephanides I agree, just made that change.

PR #99, opened 2020-02-07
GeoJSON output

brockfanning:


sdg/outputs/OutputGeoJson.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def build(self, language=None):
+        """Write the GeoJSON output. Overrides parent."""
+        status = True
+
+        target_folder = os.path.join(self.output_folder, 'geojson', self.output_subfolder)
+        if not os.path.exists(target_folder):
+            os.makedirs(target_folder, exist_ok=True)
+
+        for inid in self.indicators:
```
SavvasStephanides: Does `inid` stand for indicator ID? If so, it might be more readable to name the loop variable `indicator_id` instead. I'd argue `self.indicators` should be named `self.indicator_ids`.

sdg/outputs/OutputGeoJson.py
```diff
@@ -0,0 +1,263 @@
...
+            return False
+        # Make sure the indicator has a geocode column.
+        cols = indicator.data.columns
+        if self.id_column not in cols:
+            return False
+        # Make sure that column has data in it.
+        return indicator.data[self.id_column].any()
+
+
+    def build(self, language=None):
```
SavvasStephanides: Some splitting up of this function might be required in this function to avoid nested `if`s and `for`s

sdg/outputs/OutputGeoJson.py
```diff
@@ -0,0 +1,263 @@
...
+        for inid in self.indicators:
+            if self.indicator_has_geocodes(self.indicators[inid]):
+                no_indicators_have_geocodes = False
+                break
+
+        if no_indicators_have_geocodes:
+            print('No indicators have data under "' + self.id_column + '".')
+            status = False
+
+        return status
```
SavvasStephanides: This could be separated in its own function `geocodes_exist` (or something similar)

tests/test_build.py
```diff
@@ -38,7 +38,20 @@ def test_build(test_site_dir):
...
     assert opensdg_output.execute()
 
-    exp_dirs = set(['comb', 'data', 'edges', 'headline', 'meta', 'stats', 'zip', 'translations'])
+    # Also generate GeoJSON output.
+    geojson_output = sdg.outputs.OutputGeoJson(
+        inputs=inputs,
+        schema=schema,
+        output_folder=site_dir,
+        translations=translations,
+        geojson_file='https://geoportal1-ons.opendata.arcgis.com/datasets/4fcca2a47fed4bfaa1793015a18537ac_4.geojson',
```
SavvasStephanides: Typically not a good idea to have a remote source for data in your tests. It would potentially break if there's a problem with your internet connection. Could you download this locally and test that instead?

sdg/outputs/OutputGeoJson.py
```diff
@@ -0,0 +1,263 @@
...
+            For example, maybe a "Region" column exists with the names of the
+            regions as values. This can be used to "map" those region names to
+            geocodes, and save you the work of maintaining a separate id column.
+        """
+        if translations is None:
+            translations = []
+        if exclude_columns is None:
+            exclude_columns = []
+        if id_replacements is None:
+            id_replacements = {}
```
SavvasStephanides: Possible to initialise these in the constructor parameters?

sdg/outputs/OutputGeoJson.py
brockfanning: Agreed, I've tried to un-nest it a bit.

sdg/outputs/OutputGeoJson.py
brockfanning: I would have thought so, and tried that at first. Apparently that's a Python gotcha. Method parameters should never be given defaults that are mutable types, like lists and dicts. If you do that, Python treats them like static "class properties", and every instance of the class shares the same value. Took me a lot of debugging to figure that out.

tests/test_build.py
brockfanning: Good point, I've downloaded and committed the geojson to the repo.

sdg/outputs/OutputGeoJson.py
brockfanning: Agreed, I made that change.

sdg/outputs/OutputGeoJson.py
brockfanning: Yes, I've updated the variables names added some methods to make this more clear. `self.indicators` is actually a dict, with the indicator ids keyed to `Indicator` objects. So I've added methods `get_indicator_ids` and `get_indicator_by_id` which would make things more clear.

sdg/outputs/OutputGeoJson.py
```diff
@@ -0,0 +1,263 @@
...
+from sdg.outputs import OutputBase
+
+class OutputGeoJson(OutputBase):
+    """Output SDG data/metadata in GeoJson disaggregated by region."""
+
+
+    def __init__(self, inputs, schema, output_folder='_site', translations=None,
+        geojson_file='regions.geojson', name_property='name', id_property='id',
+        id_column='GeoCode', output_subfolder='regions', filename_prefix='indicator_',
+        exclude_columns=None, id_replacements=None):
```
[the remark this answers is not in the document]
brockfanning: 1. Agreed, I've made a change to that effect.
2. I agree for languages like Javascript that don't support named parameters. But Python does support "keyword arguments" which means that the order of those arguments doesn't matter, and the function calls are self-documenting because they're named. I think that as long as we're taking advantage of Python's keyword arguments, using an argument object would be redundant.

sdg/open_sdg.py
```diff
@@ -47,13 +47,16 @@ def open_sdg_build(src_dir='', site_dir='_site', schema_file='_prose.yml',
...
+        map_layers: list. A list of dicts describing geojson to process
         reporting_status_extra_fields: list. A list of extra fields to generate
           reporting stats for.
         config: str. Path to a YAML config file that overrides other parameters
 
     Returns:
         Boolean status of file writes
     """
+    if map_layers is None:
+        map_layers = []
```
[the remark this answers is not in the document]
brockfanning: See my other comment about why we can't do that.

sdg/open_sdg.py
```diff
@@ -47,13 +47,16 @@ def open_sdg_build(src_dir='', site_dir='_site', schema_file='_prose.yml',
         schema_file: str. Location of schema file relative to src_dir
         languages: list. A list of language codes, for translated builds
         translations: list. A list of git repositories to pull translations from
+        map_layers: list. A list of dicts describing geojson to process
```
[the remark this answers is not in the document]
brockfanning: Agreed, change made.

brockfanning:
@SavvasStephanides Thanks for the notes - I made the updates and it's ready for review again.

sdg/outputs/OutputGeoJson.py
SavvasStephanides: Makes sense

tests/test_build.py
SavvasStephanides: As a rule of thumb, unit tests should be fast, predictable and reliable. They should only fail if there's a bug in your code. If your code typically fetches data from a remote source, a good way to test this would be the usage of "mocks" or "test doubles" (in PyTest, it's called [monkeypatch](https://docs.pytest.org/en/latest/monkeypatch.html))

PR #102, opened 2020-02-11
Allow CSV metadata input

LucyGwilliamAdmin:


brockfanning:
@LucyGwilliamAdmin It looks like Travis is having some unrelated problem in its builds. Hopefully it works itself out.

In the meantime a question: I was wondering about the possibility of making the `metadata_mapping` an optional thing. If I understand correctly, if the user puts field keys in the CSV file, the metadata_mapping would not be necessary. Do I have that right? If so, I could imagine some users that prefer to just use the field keys as an alternative to maintaining a mapping file.

Maybe it could be an optional parameter, where the value of the parameter is the name of the file? Like:
```
def __init__(self, path_pattern='', git=True, metadata_mapping=None):
```

I do like the mapping file idea, and I think it would be a good optional feature to add to the YAML metadata input as well.

brockfanning:
@LucyGwilliamAdmin In the `__init__` method you'll want to put `self.metadata_mapping = metadata_mapping`, and then later in the `execute` method whenever you need to get the mapping, you can refer to `self.metadata_mapping`.

LucyGwilliamAdmin:
Thanks @brockfanning - I'm not used to the oop approach

LucyGwilliamAdmin:
@brockfanning I was still testing using previous branch with similar stuff, but I've just pointed at this branch instead and Travis checks still aren't passing:
https://travis-ci.com/open-sdg/sdg-build/jobs/286369454

Saying something about tests

brockfanning:
I'm not sure why that "pip" command is failing in Python 3.7. I can give it a try locally and look for any clues.

LucyGwilliamAdmin:
@brockfanning ok, great thanks!

I'm trying to have two separate metadata files here in [my test repo](https://github.com/LucyGwilliamAdmin/sdg-data-quick-start) - a csv which contains some fields and a md which contains some other fields.

Should this work? For some reason it's not - only the md metadata is building

brockfanning:
I would have thought that would work, but maybe one is overwriting the other. To confirm maybe you could switch the order to the CSV metadata input is last - I bet the result will be that only the CSV metadata gets imported. That would point to an issue in sdg-build we'll need to fix.

LucyGwilliamAdmin:
@brockfanning I've just checked my mapping and it looks like I saved it as wrong format so all Cyrillic characters were showing as ?.

Just re-uploaded fixed version and metadata from both files are building now :)

LucyGwilliamAdmin:
Actually I thought both metadata files were building but they're not - you're right, one must be being overwritten as only the second one in the output.py file gets built

brockfanning:
@LucyGwilliamAdmin Looking at our code, yes, it assumes that each input should have the complete data or the complete metadata. I just pushed a commit to this branch that might work. Can you give it another try?

brockfanning:
@LucyGwilliamAdmin Looks like the tests are passing now - no idea why but switching to Python 3.7 instead of Python 3.7-dev seems to have worked.

LucyGwilliamAdmin:
Great! Both csv and md metadata are building now!

Thanks for fixing the fails as well!

brockfanning:
@LucyGwilliamAdmin I meant to ask if the part about the source_* fields was necessary. The reason I ask is that, in theory, this input is not Open SDG-specific. So it would be unusual to add in code that is specific to the metadata Open SDG uses.

If it's necessary, I wonder if there is some other way to handle it, that would allow us to merge this without open-sdg-specific code. Maybe through the `add_meta_alteration` method?

brockfanning:
I'll go ahead and merge this, and follow-up about the source_* code.

LucyGwilliamAdmin:
@brockfanning I was actually thinking something re source fields - happy for that to be removed.

PR #103, opened 2020-02-13
Save json info about the zip file

brockfanning:
This saves some info (last commit date, and file size in bytes) about the exported zip files, in json form.

### DIFF
--- a/sdg/IndicatorExportService.py
+++ b/sdg/IndicatorExportService.py
@@ -1,4 +1,6 @@
 import os
+import git
+import json
 from zipfile import ZipFile
 
 class IndicatorExportService:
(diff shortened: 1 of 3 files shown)


sdg/IndicatorExportService.py
```diff
@@ -63,4 +66,26 @@ def __create_zip_file(self, zip_file_name, files_to_include):
         for each_file in files_to_include:
             zip_file.write(each_file["path"], each_file["file_name"])
 
-        zip_file.close()
\ No newline at end of file
+        zip_file.close()
+
+        self.__save_zip_file_info(zip_file_name)
+
+    def __save_zip_file_info(self, zip_file_name):
```
SavvasStephanides: I think this function could be split up / refactored a bit. Something along the lines of:

```python
def __save_zip_file_info(zip_file_name):
    info = __get_zip_file_info(zip_file_name)
    # generate the file from info

def __get_zip_file_info(zip_file_name):
    file_info = {}

    file_size = __get_file_size(zip_file_name)
    file_info["size_in_bytes"] = file_size
    file_info["size_human"] = humanize.naturalsize(file_size)  

    if __data_is_stored_in_git(): # <-- not entirely sure what this check is for, this name is an assumption
        last_commit_timestamp = __get_last_commit_timestamp()
        file_info["timestamp"] = 123

    return file_info

    # Define the rest of the functions below
````

brockfanning: @SavvasStephanides Thanks for the feedback, just pushed up a refactor.

PR #113, opened 2020-02-25
Fix logic of all-numeric check in stats.py

brockfanning:
The previous logic here either a) wasn't working as intended, b) fails in a particular edge case, or c) is broken by the upgrade of a dependency. I haven't investigated enough to know which, but in any case, it will need to be adjusted. So this PR will go ahead and adjust it, to save the time of investigating.

### DIFF
--- a/sdg/stats.py
+++ b/sdg/stats.py
@@ -43,7 +43,7 @@ def reporting_status(schema, all_meta, extra_fields=None):
     meta_df = pd.DataFrame(rows, index=all_meta.keys())
 
     # Make sure that numeric columns are numeric.
-    def value_is_numeric(val):
+    def value_is_numeric(value):
         return value if isinstance(value, int) else (isinstance(value, str) and value.isnumeric())
     for field in grouping_fields:
         if all(value_is_numeric(x) for x in meta_df[field]):
--- a/sdg/stats.py
+++ b/sdg/stats.py
@@ -43,8 +43,8 @@ def reporting_status(schema, all_meta, extra_fields=None):
     meta_df = pd.DataFrame(rows, index=all_meta.keys())
 
     # Make sure that numeric columns are numeric.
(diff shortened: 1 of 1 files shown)


sdg/stats.py
```diff
@@ -43,8 +43,10 @@ def reporting_status(schema, all_meta, extra_fields=None):
     meta_df = pd.DataFrame(rows, index=all_meta.keys())
 
     # Make sure that numeric columns are numeric.
+    def x_is_numeric(x):
```
SavvasStephanides: I would suggest not naming this function `x_is_numeric`.
```suggestion
    def value_is_numeric(value):
```

Also, do we need to keep decimal values into account? If so , Python's `isnumeric()` returns `False` when the value is decimal (for example `55.1`)

brockfanning: @SavvasStephanides Thanks, I changed that function name/parameter.

In this case I don't think we need to worry about floats. This bit of code would be used to group indicators into different categories, according to the values of metadata fields. That integer stuff was to support countries that might be using integers to categorize indicators. (Eg, 1 = national, 2 = global, etc.) In this case, we just wanted to make sure that the data type was consistent, so that all the 1's would be grouped together (for example) in case somehow Python was reading some 1's as strings and other 1's as integers. I guess that could happen if, in the YAML, there was inconsistency between `1` and `'1'`. I don't think that anyone would be using floats to categorize indicators. (To be honest, I can't remember whether that bit of code was a response to an actual problem, or just a "maybe this will be a problem" kind of thing.)

brockfanning:
@SavvasStephanides Took me a few tries but it's ready for review again. :)

sdg/stats.py
SavvasStephanides: OK. Thanks for clarifying!

PR #119, opened 2020-03-11
Attempt better Windows support

brockfanning:


brockfanning:
@LucyGwilliamAdmin On second thought I think the callback should be moved a bit. Does this still work for you?

LucyGwilliamAdmin:
@brockfanning yes, still worked

PR #120, opened 2020-03-11
Special validation for graph_title

brockfanning:


brockfanning:
@LucyGwilliamAdmin This is going to be needed for https://github.com/open-sdg/open-sdg/pull/538

brockfanning:
@LucyGwilliamAdmin Actually  - please hold off on this. I need to think about more about the ramifications.

PR #122, opened 2020-03-14
Change the Open SDG config syntax to support non-standard inputs

brockfanning:
Fixes https://github.com/open-sdg/open-sdg/issues/540

This adds a new option to the config-file approach for Open SDG: `inputs`. Documentation of the new option is in the PR.

This also has a breaking change related to the `translations` option. (Though since the config-file approach hasn't been in an official release, it's not so bad.)

The breaking change is that the `translations` config setting is no longer a list of strings. Now it is a list of dicts. For example, instead of:

```
translations:
  - https://github.com/open-sdg/sdg-translations.git@master
```

It is now:

```
translations:
    source: https://github.com/open-sdg/sdg-translations
    tag: master
```

This is bit more technical, but allows for including the other types of translations (notably SDMX translations).


brockfanning:
@LucyGwilliamAdmin This is my best try at allowing the config-file approach to be used for SDMX or other non-standard inputs.

The problem is that I don't think this is feasible for Kyrgyzstan, because they are using a whole folder of SDMX files. There's no way to "loop" in a YAML file. We would need to create a new input class (maybe `InputSdmxMultiple`?) that could scan a whole folder and import all the SDMX at once. This shouldn't be too terribly hard, and I can look into it if you think it will be needed.

brockfanning:
@LucyGwilliamAdmin Actually it wasn't too hard to do that InputSdmxMl_Multiple class. So I think once #123 is merged then this could be extended to use it, and Kyrgyzstan would be able to benefit from it. The only limitation is that the `data_alteration` stuff can't be put into a YAML file, but YAML can't contain Python code.

LucyGwilliamAdmin:
@brockfanning I'm not really sure how it works but could fix_data be specified in a custom .py file in the assets folder? similar to how we do with js and css?

brockfanning:
@LucyGwilliamAdmin Hmm, that could work. I'll do some more thinking on it.

brockfanning:
@LucyGwilliamAdmin I've given this a try but looking for a specific file: open_sdg_alter.py containing functions called "alter_data" and/or "alter_meta". I've got some explanation in the examples in this PR. I did some quick testing but could use your input.

LucyGwilliamAdmin:
@brockfanning I'm just testing this out, and I'm getting an error in Actions:
https://github.com/sdg-kyrgyzstan/open-sdg-data-starter/runs/516299455?check_suite_focus=true

If I'm using SDMX and Excel input do I need to also list the CSV data input class and the Yaml MD input class under inputs?

LucyGwilliamAdmin:
Just realised I still had some Python in the file, so have just removed that and testing

LucyGwilliamAdmin:
I've made those changes but still having some trouble:
https://github.com/sdg-kyrgyzstan/open-sdg-data-starter/runs/516397329?check_suite_focus=true

LucyGwilliamAdmin:
@brockfanning could you take a look over my file if you have chance please?
https://github.com/sdg-kyrgyzstan/open-sdg-data-starter/blob/config/open_sdg_config_sdmx.yml

brockfanning:
@LucyGwilliamAdmin Sure thing, I should have some results for you in about an hour.

brockfanning:
@LucyGwilliamAdmin There were a few issues:
1. There's a stray comma in the config file on [this line](https://github.com/sdg-kyrgyzstan/open-sdg-data-starter/blob/config/open_sdg_config_sdmx.yml#L25) that was causing the loading of the config file to fail. And unfortunately it was failing silently so it was not obvious this was happening.
2. The new InputSdmxMl_Multiple class was not yet included in this PR. Each input has to be specifically supported in the code, in [this section](https://github.com/open-sdg/sdg-build/pull/122/files#diff-958f73e64593ce4c233a3715a59c513fR239) and [this section](https://github.com/open-sdg/sdg-build/pull/122/files#diff-958f73e64593ce4c233a3715a59c513fR260).

For the "failing silently" problem I've pushed an update to this branch that will make sure to raise an exception whenever that happens in the future.

For the second problem I've pushed an update to this branch that adds support for InputSdmxMl_Multiple. Also, since you're testing with Kyrgyzstan, I've also gone ahead and added support for InputExcelMeta, even though that's not in sdg-build yet.

So I think the next steps for testing would be:
1. Merge the latest from this branch into your excel-metadata-kyrgyzstan branch
2. Remove that stray comma
3. Test again

brockfanning:
@LucyGwilliamAdmin Great! Just wanted to confirm though: are you sure that the alteration functions were working as expected? I just want to make sure it's looking in the right place.

LucyGwilliamAdmin:
@brockfanning think so, their [1-2-2 file](https://github.com/sdg-kyrgyzstan/open-sdg-data-starter/blob/develop/data/1.2.2.xml) has source_detail in but it's not showing as a disaggregation on the [platform](https://sdg-kyrgyzstan.github.io/open-sdg-site-starter/en/1-2-2/)

brockfanning:
@LucyGwilliamAdmin Sorry but can I ask you to review one more alternative approach? I realized that there is already a place where we can inject Python code: the place where `open_sdg_build` and `open_sdg_check` are called. So I've tried adding `alter_data` and `alter_meta` parameters to those functions. I added [examples of how this would be used](https://github.com/open-sdg/sdg-build/pull/122/files#diff-efe72952580d8fd695f3e23fc8a6d82eR17-R23).

If it works, I think this might be a bit better, as it has less uncertainty and requires less documentation.

LucyGwilliamAdmin:
@brockfanning I have just tried this way on Kyrgyzstan site and seems to still be working - also fixes issue I mentioned briefly the other day about being able to set the same metadata field across all indicators.

LucyGwilliamAdmin:
I ended up combining their check and build scripts just so code didn't have to be repeated:
https://github.com/sdg-kyrgyzstan/open-sdg-data-starter/blob/develop/scripts/check_and_build_data.py

brockfanning:
@LucyGwilliamAdmin That looks great - and it still worked OK?

brockfanning:
Ah, missed your first note. Cool! I will merge.

PR #123, opened 2020-03-14
SDMX-ML multiple input

brockfanning:
This should allow for input from a folder of SDMX files, as a more convenient alternative to looping through files in Python. It checks to see if the file is Structure or StructureSpecific.


brockfanning:
@LucyGwilliamAdmin This may be of interest for the Kyrgyzstan site. Here's what it might be used there:

```
sdmx_pattern = os.path.join('data', '*.xml')
sdmx_input = sdg.inputs.InputSdmxMl_Multiple(
    path_pattern=sdmx_pattern,
    import_translation_keys=True,
    dimension_map=dimension_map,
    dsd=dsd,
    drop_dimensions=drop_dimensions,
    data_alterations=[fix_data],
)
inputs.append(sdmx_input)
```

This should be all you need to do for the SDMX input - so you could remove the looping through the SDMX files, and the testing for what type of SDMX file each one is. That's all handled inside this new input.

PR #125, opened 2020-03-15
Search inside Git repo for file paths - for better Windows Cygwin support

brockfanning:
When run in Windows, the Windows paths (with backslashes) don't show up as being in the repo, if the repo was created in non-Windows (with forward slashes). This causes exceptions. This is a different way at getting to the needed commit, just using the repo paths only, so hopefully Windows won't be able to cause a problem.

EDIT: On further testing I see that this is only necessary when using Cygwin in Windows (or Cygwin wrappers like Babun). So it's a fairly narrow edge case. Probably it's a good idea anyway, assuming it doesn't break anything. But I would be fine postponing this until after version 1.

### DIFF
--- a/sdg/git.py
+++ b/sdg/git.py
@@ -13,17 +13,17 @@ import git
 from sdg.path import input_path  # local package
 
 # %% Faster but not using right now
-# 
+#
 # def get_last_update(obj, repo):
 #     commit = next(repo.iter_commits(paths=obj.path, max_count=1))
 #     git_date = str(commit.committed_datetime.date())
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
@brockfanning this was such a long time ago I can't remember what issue I was having - is this still something worth doing?

brockfanning:
@LucyGwilliamAdmin I would be fine closing this if you agree.

LucyGwilliamAdmin:
Fine with me also

PR #126, opened 2020-03-16
Add Excel metadata input and make CSV metadata input consistent

LucyGwilliamAdmin:


LucyGwilliamAdmin:
@brockfanning slight side note, when I was pointing at a diff branch for Kyrgyzstan the [Excel Meta input script](https://github.com/open-sdg/sdg-build/blob/68ca0addf8124599ff904b8cff19f8d35b7fca28/sdg/inputs/InputExcelMeta.py) was a little more precise e.g. as they had already created the files the naming was **Метаданные [INDICATOR-CODE].xlsx** so I included [this line](https://github.com/open-sdg/sdg-build/blob/68ca0addf8124599ff904b8cff19f8d35b7fca28/sdg/inputs/InputExcelMeta.py#L52) at the bottom to make the script work properly `inid = inid.split()[1].replace(".","-")`

Now that the script is more generalised, is there anything I can do to replicate this?

brockfanning:
@LucyGwilliamAdmin That's a tough one. One way to do that would be to make your own sub-class and to override [this method](https://github.com/open-sdg/sdg-build/blob/master/sdg/inputs/InputFiles.py#L17). But then you would not be able to use the config file approach.

We could also try to figure out how to add this kind of thing to the "open_sdg_alter" stuff from the other PR. But that seems like something that would need some careful consideration and maybe some refactoring.

Do you think it would be possible to require a certain naming convention for the files?

sdg/inputs/InputExcelMeta.py
```diff
@@ -0,0 +1,66 @@
...
+            if inid is not None:
+                fr = os.path.join(path, inid + '.xlsx')
+            else:
+                fr = path
+            meta_excel=pd.ExcelFile(fr)
+            meta_df=meta_excel.parse(meta_excel.sheet_names[sheet_number])
+            for index, row in meta_df.iterrows():
+                if type(row[0])==float:
+                    if np.isnan(row[0]):
+                        meta_df.iat[index-1,1]=row[1]
```
brockfanning: Would it be possible to add a bit of commenting here to explain what's going on in these lines?

sdg/inputs/InputExcelMeta.py
```diff
@@ -0,0 +1,66 @@
...
+                        meta_df.iat[index, 1]=np.nan
+            meta_df=meta_df.dropna()
+            meta_df.columns=["Field name", "Field key"]
+            if metadata_mapping != None:         
+                meta_mapping_df=pd.merge(meta_mapping, meta_df, on="Field name")
+                meta=dict()
+                for row in meta_mapping_df.iterrows():
+                    if type(row[1][2])==float:
```
brockfanning: Could you add some commenting here to explain what's going on? In particular, what are `row[1][1]`, `row[1][2]`, etc referring to.

LucyGwilliamAdmin: Looking at this again line 40 to 44 here is specific to Kyrgyzstan metadata files so I'm removing those lines now

sdg/inputs/InputExcelMeta.py
LucyGwilliamAdmin: Looking at this again line 40 to 44 here is specific to Kyrgyzstan metadata files so I'm removing those lines now

sdg/inputs/InputExcelMeta.py
LucyGwilliamAdmin: The reason these if-else statements exists is because NaN was causing an error, although now I'm thinking about it the NaNs should be removed by the code above. I'll take another look and let you know when it's ready for review again

LucyGwilliamAdmin:
@brockfanning think this is ready for review again

LucyGwilliamAdmin:
@brockfanning re. naming convention - I guess that would make sense. I'll add something to the documentation, and speak to Kyrgyzstan about changing the name of their files. I'll also speak to them to see if they can restructure their files slightly so that the other part of specific code isn't needed.

PR #128, opened 2020-03-16
Reporting_status also to default to complete if data_non_statistical is true

LucyGwilliamAdmin:


LucyGwilliamAdmin:
@brockfanning yes, I'll add them back in - not sure what happened as don't think I removed them manually

PR #135, opened 2020-03-19
Allow translations to overrides previous ones

### LINKED ISSUE #118: Subsequent translation inputs don't override previous ones
Right now, if you have multiple translation inputs, the first ones take precedence over the later ones.

For example, if you import translations from a remote Git source, in which the Spanish translation for "my_group.foo" is equal to "bar", and then afterwards you import translations from a local folder of YAML files, in which the Spanish translation for "my_group.foo" is equal "baz", the end result if that it will be "bar".

I think this is not what one would expect, and we should change it so that the later inputs override the earlier ones. I believe the problem code is [here](https://github.com/open-sdg/sdg-build/blob/master/sdg/translations/TranslationInputBase.py#L70). In theory we could remove that conditional. I don't believe it would have any negative consequences, but testing would be helpful.

brockfanning:
Trying this again.

Fixes #118

### DIFF
--- a/sdg/inputs/InputCsvMeta.py
+++ b/sdg/inputs/InputCsvMeta.py
@@ -50,7 +50,6 @@ class InputCsvMeta(InputFiles):
                 # Loop through dataframe rows, assigning first column item to dictionary key
                 # and second column item to dictionary value
                 for row in meta_csv.iterrows():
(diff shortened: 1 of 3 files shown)


brockfanning:
@LucyGwilliamAdmin Here's another try at that PR.

PR #139, opened 2020-03-26
Preserve the order of the fields in the schema input/output

### LINKED ISSUE #138: Random order of fields in schema output
The order of the fields in the schema output does not preserve the order of the fields in the input. I guess this is because it's converted to a dict, which doesn't preserve order. If possible we should try to preserve the order. Otherwise platforms like Open SDG, which rely on the order, will always display metadata fields in a random order.

brockfanning:
Fixes #138

### DIFF
--- a/sdg/schemas/SchemaInputBase.py
+++ b/sdg/schemas/SchemaInputBase.py
@@ -173,7 +173,7 @@ class SchemaInputBase:
         return {x['enum'][0]: x['translation_key'] for x in options}
 
 
-    def add_field_order(self, field):
+    def add_item_to_field_order(self, field):
         """Add a field to the list, in case an output needs a field order.
 
         Parameters
@@ -192,6 +192,4 @@ class SchemaInputBase:
         list
             A list of field names in a particular order
         """
-        if len(self.field_order) == 0:
-            raise Exception('The schema input did not preserve the order of fields.')
-        return self.field_order
+        return self.field_order if len(self.field_order) > 0 else self.schema['properties'].keys()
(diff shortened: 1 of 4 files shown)


brockfanning:
@SavvasStephanides @LucyGwilliamAdmin This is a regression bug that I think should be fixed before version 1. The bug is that every time a build happens, the metadata fields get shuffled into a random order. This PR fixes the problem for me locally.

sdg/schemas/SchemaInputBase.py
```diff
@@ -170,3 +171,27 @@ def get_value_translation(self, field):
             raise ValueError(field + " field does not have options element")
 
         return {x['enum'][0]: x['translation_key'] for x in options}
+
+
+    def add_field_order(self, field):
```
SavvasStephanides: Consider renaming this function to something that makes it clear that you're adding something to the field_order list. `add_item_to_field_order` or similar?

sdg/schemas/SchemaInputBase.py
```diff
@@ -170,3 +171,27 @@ def get_value_translation(self, field):
...
+    def get_field_order(self):
+        """Get the list of fields in the preserved order.
+
+        Returns
+        -------
+        list
+            A list of field names in a particular order
+        """
+        if len(self.field_order) == 0:
+            raise Exception('The schema input did not preserve the order of fields.')
```
SavvasStephanides: Maybe the function should default to something instead of throwing an exception?

brockfanning: Good point, it can just default to the keys. They'll be in a random order, but I guess that's OK.

sdg/schemas/SchemaInputBase.py
brockfanning: Thanks, just renamed it. As for why, just a general habit of trying to use methods instead of directly working with properties. I don't always remember to do that, but I try. No particular reason other than that.

PR #140, opened 2020-04-03
Print warning instead of exception when series has duplicate value

brockfanning:
This allows builds to complete even when there are issues in the data.

### DIFF
--- a/sdg/Indicator.py
+++ b/sdg/Indicator.py
@@ -334,7 +334,7 @@ class Indicator:
             serialized = json.dumps(disaggregations, sort_keys=True)
             # Initialized any new series.
             if serialized not in all_series:
-                all_series[serialized] = sdg.Series(disaggregations)
+                all_series[serialized] = sdg.Series(disaggregations, self.get_indicator_id())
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin:
@brockfanning this is great - much easier to see where the issues are - wondering if the indicator code could also be in the warning?

LucyGwilliamAdmin:
Here's how warning looks:
https://travis-ci.org/github/ONSdigital/sdg-data

brockfanning:
@LucyGwilliamAdmin Thanks - any feedback? Is it helpful enough?

brockfanning:
@LucyGwilliamAdmin Sorry, missed earlier comment. I'll see about the indicator code, though I think that might be tricky.

LucyGwilliamAdmin:
No problem if not - I'm mainly just thinking about if someone is bulk uploading.

I'm sure there's enough info for someone to be able to work what indicator the series belongs to!

brockfanning:
@LucyGwilliamAdmin I just pushed up an attempt - can you give it a try?

PR #146, opened 2020-04-24
Do not translate 'Year', 'Value', or 'Units' in data column names

### LINKED ISSUE #145: Translated builds fail if special columns are translated
If the special columns ("Year", "Units", and "Value") have translations, during the translated build process, an error occurs, because some code expects those columns to be there, in English. More investigation is needed.

brockfanning:
Fixes #145


brockfanning:
This fixes a bug, so I'm going to go ahead and merge, and tag a hotfix.

PR #150, opened 2020-05-31
Don't omit the SERIES ids from disaggregations in SDMX inputs

brockfanning:
This is untested but may help with the Series column feature.

### DIFF
--- a/sdg/inputs/InputSdmxJson.py
+++ b/sdg/inputs/InputSdmxJson.py
@@ -153,8 +153,8 @@ class InputSdmxJson(InputSdmx):
         for key in dimensions:
             dimension = dimensions[key]
             dimension_id = dimension['dimension']['id']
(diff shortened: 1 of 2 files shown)


brockfanning:
@LucyGwilliamAdmin It looks like this works to get the series into the build. It shows up as the human-readable label of the series. Is that what you would want? If so, I think this is ready for review. Or would you want the series id instead?

brockfanning:
@LucyGwilliamAdmin Correction, I guess it's actually translating them. So I think that's correct, and this is ready for review.

LucyGwilliamAdmin:
Nazira has also confirmed that she is happy with the feature. She did ask about whether it would be possible to select more than one series at a time - so I guess it would act more like a disaggregation than a Units type but think we could create new issue for that?

brockfanning:
@LucyGwilliamAdmin Thanks for the testing! I'll follow up on that question over at https://github.com/open-sdg/open-sdg/pull/678

PR #153, opened 2020-06-01
Automatic output documentation

### LINKED ISSUE #147: Produce HTML page documenting the output
For the purposes of documenting APIs and download locations, implementations could benefit from an HTML page with human-readable examples of each output. I imagine that we would need to add some code to each output with a method for generating HTML.

This is rough example of what I'm thinking of:

---

# Examples of output

## Notes

* In the examples below, the language code `en` can be replaced as needed.
* In the examples below, the indicator number `1-1-1` can be replaced as needed.

## Data

* CSV: /en/data/1-1-1.csv
* JSON: /en/data/1-1-1.json

## Metadata

* JSON: /en/meta/1-1-1.json

## GeoJSON

GeoJSON data is available in the following layers:

* /en/geojson/country/indicator_1-1-1.geojson
* /en/geojson/regions/indicator_1-1-1.geojson

GeoJSON data is available for the following indicators: 1-1-1, 1-2-1, 1-3-1

## Zip

A zip of all CSV files is available here:

* /en/zip/all_indicators.zip

brockfanning:
Fixes #147 

This can be customized with the `docs_branding` and `docs_intro` config options, but otherwise should be mostly automatic. Builds generated should automatically get an HTML homepage with info about the various outputs.


brockfanning:
@LucyGwilliamAdmin @SavvasStephanides This is very basic but it's a start. Feedback is welcome. The idea is to give the data repository a homepage with links to some examples of the various endpoints.

You can see it in action here: http://brock.tips/sdg-data-kazstat-upgrade/

brockfanning:
Actually hold off on reviewing, I've got some ideas for improvement.

brockfanning:
@LucyGwilliamAdmin @SavvasStephanides Ok, this is ready for review again. :)

sdg/DocumentationService.py
```diff
@@ -0,0 +1,159 @@
+import os
+import sdg
+import numpy
+from slugify import slugify
+
+class DocumentationService:
```
SavvasStephanides, line 6: Great that you prefixed your class name with -service. Makes it clear that it performs an action. However, `DocumentationService` is a bit vague. Consider `OutputDocumentationService`, `BuildDocumentationService` or `BuildOutputDocumentationService`. I believe it would make it slightly clearer of what the purpose of this service is.

sdg/DocumentationService.py
```diff
@@ -0,0 +1,159 @@
...
+    """HTML generation to document outputs built with this library.
+
+    Note that this is meant to document particular builds, not the library in
+    general. The idea is that each time this library is used to build/convert
+    some SDG-related data, this class can be used to generate human-friendly
+    HTML pages documenting the specifics of the build (such as endpoint URLs).
+    """
+
+
+    def __init__(self, outputs, folder='_site', branding='Build docs', languages=None, intro=''):
```
SavvasStephanides, line 16: Wouldn't the documentation generation require at least one language to work? If so (I might be missing something), you might want to inject this dependency on your constructor (omit `=None`), or you could default it to `languages=['en']`

sdg/DocumentationService.py
```diff
@@ -0,0 +1,159 @@
...
+            title = output.get_documentation_title()
+            content = output.get_documentation_content(self.languages)
+            pages.append({
+                'title': title,
+                'filename': self.create_filename(title),
+                'content': content
+            })
+
+        if not os.path.exists(self.folder):
+            os.makedirs(self.folder)
```
SavvasStephanides, line 45: `os.makedirs` has a [built-in check](https://docs.python.org/3/library/os.html#os.makedirs) for checking if the folder already exists, if you'd like to do without the `if` statement (`exist_ok=True`)

sdg/DocumentationService.py
```diff
@@ -0,0 +1,159 @@
...
+
+        for page in pages:
+            self.write_documentation(page)
+
+        self.write_index(pages)
+
+
+    def write_page(self, filename, html):
```
SavvasStephanides, line 53: Best if you define your functions after they're called, in the order they are called (the newspaper paradigm we discussed before).

sdg/DocumentationService.py
brockfanning: Good idea, just changed it to OutputDocumentationService.

sdg/DocumentationService.py
brockfanning: No actually it should be able to work without a language. In sdg-build we have this concept of "translated builds", where you specify one or more languages, and if you do, it creates entire builds for each of those languages, in subfolders. But this is not required.

sdg/DocumentationService.py
brockfanning: Cool, just made that change.

sdg/DocumentationService.py
brockfanning: Makes sense.

brockfanning:
@SavvasStephanides Thanks for the feedback, changes pushed.

sdg/OutputDocumentationService.py
```diff
@@ -0,0 +1,162 @@
...
+        outputs : list
+            List of objects inheriting from OutputBase
+        """
+        self.outputs = outputs
+        self.folder = folder
+        self.branding = branding
+        self.intro = intro
+        self.slugs = []
+        self.languages = [] if languages is None else languages
+
```
SavvasStephanides, line 30: The order your functions are being called:

* `generate_documentation`
* `create_filename`
* `write_documentation`
* `write_index`
* `get_html`
* `write_page`

It'll probably make your code more readable if you define your functions in this order.

SavvasStephanides:
@brockfanning Maybe I rushed to approve this review. I noticed that there are a lot of empty `json` and `csv` files (especially under Edges). Is this OK or should be have a filter in place to prevent these, since we're showing examples?

Also, I'll wait until @LucyGwilliamAdmin points out any issues.

brockfanning:
@SavvasStephanides That's a good question. The code arbitrarily picks 3 random indicators to use. We could add code that keeps cycling through the indicators until it finds 3 that have both headlines and edges. However we'd still need to allow for the possibility that there are no indicators that meet those requirements, and in that case it would need to still fallback to showing the original 3 random indicators.

One compromise to all that code could be the ability to optionally configure which indicators are used. Like, a `docs_examples` config setting.

Lastly an argument could be made that it's actually good to show examples with nothing, because that does show how the endpoint will behave when there is no data.

brockfanning:
@SavvasStephanides Actually, it wasn't too hard to do those checks, with a fallback, and I think cleans it up a bit too. The PR and example have been updated.

LucyGwilliamAdmin:
@brockfanning @SavvasStephanides I am happy with this. My only thoughts are should we provided some sort of link between this and the platform e.g. a link on platform, and a link on data site?

brockfanning:
There's already a `docs_intro` config setting where users to add links to their platform. Does that work? That's what I've done in the linked example.

For the platform side, this would be pretty easy. The URL is already available to Jekyll at `site.remote_data_prefix`.

LucyGwilliamAdmin:
Ahh great, yes that works

PR #154, opened 2020-06-03
Rem input path

brockfanning:
@LucyGwilliamAdmin Did you want to merge in this feature or are you still testing it?

### DIFF
--- a/sdg/meta.py
+++ b/sdg/meta.py
@@ -36,7 +36,7 @@ def read_meta(inid, path_pattern='', git=True, src_dir='', git_data_dir=None):
 
     # Now look for all subfolders of the meta folder, which may contain
     # multilingual metadata, and add them as well.
-    languages = next(os.walk(meta_folder))[1]
+    languages = next(os.walk(os.path.join(src_dir, meta_folder)))[1]
     for language in languages:
         i18n_fr = os.path.join(src_dir, meta_folder, language, inid + extension)
         if os.path.isfile(i18n_fr):
--- a/sdg/meta.py
+++ b/sdg/meta.py
@@ -19,7 +19,7 @@ def read_meta(inid, path_pattern='', git=True, src_dir='', git_data_dir=None):
         meta_folder = 'meta'
         extension='.md'
     else:
-        meta_folder = path_pattern.split("/")[0]
+        meta_folder = os.path.split(path_pattern)[0]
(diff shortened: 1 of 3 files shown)


sdg/meta.py
```diff
@@ -11,12 +11,20 @@
...
     """Perform pre-processing for the metadata files"""
     status = True
     # Read and write paths may be different
-    fr = input_path(inid, ftype='meta', src_dir=src_dir)
-
+    if path_pattern == '':
+        meta_folder = 'meta'
+        extension='.md'
+    else:
+        meta_folder = path_pattern.split("/")[0]
```
brockfanning, line 22: After looking a bit more, I think around here is where the problematic bits might be. A couple of issues maybe:
1. Instead of hardcoding "/" we should use `os.path.sep` (easy fix)
2. Bigger issue - this won't work if the path_pattern has multiple parts, like `path/to/my/meta/*-*.md` for example. I think maybe this is related to why the tests are failing. ([Here is the path pattern for the test suite.](https://github.com/open-sdg/sdg-build/blob/master/tests/test_build.py#L26))

LucyGwilliamAdmin: @brockfanning thanks for taking a look. What id we use os.path.split e.g. see here:
https://code.sololearn.com/crI157thO5WX/#py

Fixes hardcoding and takes all parts except for last part i.e. gets the folder paths without the file name

brockfanning: @LucyGwilliamAdmin That sounds good to me.

LucyGwilliamAdmin:
@brockfanning have just made that change but tests still failing - looks like 'tests/tests/meta/2-4-1.md' is being used as the file name

LucyGwilliamAdmin:
@brockfanning I think some of the stuff from this branch can be used, but I think a new branch needs to be created. I'm not really sure what's wrong - although it's coming back to me that this was supposed to be for data and metadata, while I've only covered the metadata stuff.

I've sort of just been playing around with it, trying to get the checks to pass but haven't had any luck - although there is a new error now, which had led me to think:

does anything need to be updated in the tests to allow for the new way of how translations are configured?

brockfanning:
@LucyGwilliamAdmin Is it possible to close this now that #170 has been merged? Or is there still something more needed?

LucyGwilliamAdmin:
@brockfanning yes, this can be closed 👍

PR #156, opened 2020-06-23
More efficient generation of unique combinations of disaggregations

brockfanning:
The code that generates unique combinations of disaggregations and converts them into "Series" objects is using a loop which is causing complex builds to take a long time to complete. This refactor is intended to accomplish the exact same results but faster.

Testing this should confirm two things:
1. The build runs significantly faster
2. The build is being generated correctly

Since the part of the code this is used is in the GeoJSON generation, I would recommend spot-checking several maps to make sure the build has correct data.

### DIFF
--- a/sdg/Indicator.py
+++ b/sdg/Indicator.py
@@ -2,6 +2,7 @@ import copy
 import json
 import sdg
 import pandas as pd
+import numpy as np
 from sdg.translations import TranslationHelper
 
 class Indicator:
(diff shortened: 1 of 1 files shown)


brockfanning:
@LucyGwilliamAdmin @SavvasStephanides This is in response to your experiences with the builds taking much longer. Can you try this out and see if there is an improvement? And also confirm that map data is still correct?

LucyGwilliamAdmin:
@brockfanning I've taken a look at this.

Before moving over to v1, data repo builds were taking up to 3 minutes. After moving to v1 they were taking approx. 15 mins. The build pointing at the took just over 8 minutes, so not as quick as before but definitely an improvement on the 15 min build time.

Re. data being correct, I've checked a couple of maps and data seems fine

SavvasStephanides:
@brockfanning Tried it myself and got similar results as @LucyGwilliamAdmin (8 minutes 13 sec). I guess it's up to the Data team to decide if this is acceptable or not.

brockfanning:
@SavvasStephanides @LucyGwilliamAdmin Wonderful! I think this is worth merging then.

Regarding the build time increase from before version 1 - remember that before version 1, the builds were not generating any GeoJSON files. Now, they are generating GeoJSON files for each indicator with a "GeoCode" column. I just did a search and see that that is 59 indicators. So, the builds are generating 59 GeoJSON files now (versus 0 GeoJSON files before version 1).

FYI if I search for `data_show_map: true` I see that it is only set on 12 indicators. So, right now you're generating GeoJSON files for 47 indicators that technically don't need it. Another quick fix to speed up your build times would be to remove the GeoCode column from your data for all indicators that don't need a map. Do you think that's feasible?

brockfanning:
Actually a correction - I said that the build is generating 59 GeoJSON files - in reality it's 177 (59 times 3), because it generates one GeoJSON file per map layer, and you've got 3 map layers.

AnnCorp:
I think we should discuss this one at tech call later.  There was a clear direction given to us to include geocodes in data where they are available, so don't think we would want to delete them. Perhaps we need to only build geojson where data_show_map is true? Or actually is having the geojson useful for others for mapping even if we don't have a map, so we'd have to accept longer build times? We do need to review though where don't have maps but have geocodes in the data.

brockfanning:
I have a feeling that the "spirit" of that policy (to include geocodes where available) would imply that having the GeoJSON available is desired, even if you are not using them on the Open SDG platform for maps. In that case you'd continue as-is, and accept the long build times. The only caveat there is that the GeoJSON files are not being "advertised" anywhere. (That gave me an idea for a feature request: #157)

PR #158, opened 2020-06-26
Add a disaggregation report to automated docs

### LINKED ISSUE #155: Add disaggregation report to Open SDG documentation
The automated documentation of the Open SDG output could benefit from a report of the disaggregations being used in data. This would be helpful in preparing for SDMX output. More thought is needed, but I wanted to get this in writing before I forgot about it.

brockfanning:
Fixes #155

### DIFF
--- a/docs/examples/open_sdg_config.yml
+++ b/docs/examples/open_sdg_config.yml
@@ -98,3 +98,4 @@ docs_branding: My build documentation
 docs_intro: This is a list of examples of endpoints and output that are
   available on this service. Click each of the links below for more information
   on the available output.
+docs_indicator_url: https://my-github-org/my-site-repository/[id]
--- a/docs/examples/open_sdg_config.yml
+++ b/docs/examples/open_sdg_config.yml
@@ -91,6 +91,22 @@ map_layers:
   #     _site/en/geojson/regions/indicator_1-2-1.geojson
   # Add any number of additional layers, such as one for counties.
 
+# Indicator options
+# -----------------
+# This control optional settings to control how your indicators are loaded.
+# Below are all the possible options, along with the defaults.
+indicator_options:
+  # This specifies the columns that should not be considered disaggregations.
+  non_disaggregation_columns:
+    - Year
+    - Units
+    - Series
+    - Value
+    - GeoCode
+    - Observation status
+    - Unit multiplier
+    - Unit measure
+
 # Documentation settings
 # ----------------------
 # This controls the automatic documentation that is created for your build.
--- a/sdg/DisaggregationReportService.py
+++ b/sdg/DisaggregationReportService.py
@@ -217,7 +217,7 @@ class DisaggregationReportService:
         for value in info['values']:
             row = {
                 'Value': self.get_disaggregation_value_link(info['values'][value]),
-                'Number of instances': info['values'][value]['instances'],
+                'Disaggregation combinations using this value': info['values'][value]['instances'],
                 'Number of indicators': len(info['values'][value]['indicators'].keys()),
             }
             for language in self.get_additional_languages():
@@ -226,7 +226,7 @@ class DisaggregationReportService:
 
         columns = ['Value']
         columns.extend(self.get_additional_languages())
-        columns.extend(['Number of instances', 'Number of indicators'])
+        columns.extend(['Disaggregation combinations using this value', 'Number of indicators'])
 
         df = pd.DataFrame(rows, columns=columns)
         if not df.empty:
@@ -251,7 +251,7 @@ class DisaggregationReportService:
         for indicator_id in info['indicators']:
             rows.append({
                 'Indicator': self.get_indicator_link(indicator_id),
-                'Number of instances': info['indicators'][indicator_id]
+                'Disaggregation combinations using this value': info['indicators'][indicator_id]
             })
         df = pd.DataFrame(rows)
         if not df.empty:
--- a/sdg/DisaggregationReportService.py
+++ b/sdg/DisaggregationReportService.py
@@ -267,11 +267,12 @@ class DisaggregationReportService:
 
 
     def get_indicator_link(self, indicator_id):
+        indicator_label = '#' + indicator_id
         if self.indicator_url is None:
-            return indicator_id
-        link = '<a href="{href}">{indicator_id}</a>'
+            return indicator_label
+        link = '<a href="{href}">{label}</a>'
         href = self.indicator_url.replace('[id]', indicator_id)
-        return link.format(href=href, indicator_id=indicator_id)
+        return link.format(href=href, label=indicator_label)
 
 
     def remove_links_from_dataframe(self, df):
--- a/sdg/DisaggregationReportService.py
+++ b/sdg/DisaggregationReportService.py
@@ -101,7 +101,7 @@ class DisaggregationReportService:
         string
             The title converted into a unique *.html filename
         """
-        slug = slugify(title)
+        slug = 'disaggregation--' + slugify(title)
         if slug in self.slugs:
             slug = slug + '_'
         if len(slug) > 100:
--- a/sdg/DisaggregationReportService.py
+++ b/sdg/DisaggregationReportService.py
@@ -56,6 +56,8 @@ class DisaggregationReportService:
         for indicator_id in indicators:
             if not indicators[indicator_id].is_statistical():
                 continue
+            if not indicators[indicator_id].is_complete():
+                continue
             non_disaggregation_columns = indicators[indicator_id].options.get_non_disaggregation_columns()
             for series in indicators[indicator_id].get_all_series():
                 disaggregations = series.get_disaggregations()
--- a/sdg/DisaggregationReportService.py
+++ b/sdg/DisaggregationReportService.py
@@ -159,7 +159,8 @@ class DisaggregationReportService:
         columns.extend(['Number of indicators', 'Number of values'])
 
         df = pd.DataFrame(rows, columns=columns)
-        df.sort_values(by=['Disaggregation'], inplace=True)
+        if not df.empty:
+            df.sort_values(by=['Disaggregation'], inplace=True)
         return df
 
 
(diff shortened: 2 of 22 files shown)


brockfanning:
@LucyGwilliamAdmin Is it possible to test this in a feature branch of the UK data repo? I could post screenshots, but I thought it would be easier to click around, and I think @AnnCorp would like to see this first draft.

brockfanning:
@LucyGwilliamAdmin @AnnCorp Actually I realized I could try it with KZ. Here is an example: http://brock.tips/sdg-data-kazstat-upgrade/

I would like to add a language toggle. Right now it is only showing the untranslated raw value of the data, but it would be nice to see it translated into the available languages.

brockfanning:
Rather than a language toggle (which would be pretty complicated) and just added additional columns in the table for the non-default languages.

LucyGwilliamAdmin:
@brockfanning have just created feature branch for UK - getting NaN for every "Number of indicators" and Number of values"

brockfanning:
@LucyGwilliamAdmin Thanks, I had a looping problem. Can you give it another try?

LucyGwilliamAdmin:
@brockfanning looks great: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/disagg-report/disaggregations/

brockfanning:
@LucyGwilliamAdmin Great - I also just added documentation of an optional feature. You can add "docs_indicator_url" to your data config file. If provided, the "[id]" will be replaced with the indicator id (dash-delimited). For example, "https://example.com/[id].html" will be replaced with "https://example.com/4-1-1.html". This will let the indicators actually link to your site repository.

I've got one more visual tweak I'd like to make - to fix the formatting of the "cards"  on the front page.

brockfanning:
@LucyGwilliamAdmin Ok, I'm done with those formatting fixes.

brockfanning:
@LucyGwilliamAdmin @SavvasStephanides I realized that this broke the "local_data_folder" approach (where site and data are in the same repository) which is what we use for the open-sdg automated tests. So I've added the option to configure the subfolder that's used in the open-sdg data config. I've also added a quick automated test for this automated documentation functionality.

This is ready (again) for review.

LucyGwilliamAdmin:
@brockfanning @AnnCorp @SavvasStephanides 

I showed this to UK data team this morning and they had some comments

- There were a couple of fields they thought maybe shouldn't be listed: Units, Unit measure, Unit multiplier and Observation status (and I think Series, especially if `series_toggle: true`)
- Group is also showing up which is a disaggregation in the placeholder data files, so think we need to ensure only reported statistical indicators feed into report

I agree with both of the points above, but they also mentioned another point, which I think will need some more thought/exploration and that we should come back to:

- A third column which is a higher level of disaggregation e.g. in this new column, the value would be Age for the rows: Age of oldest person, Age of youngest person, Age

brockfanning:
> There were a couple of fields they thought maybe shouldn't be listed: Units, Unit measure, Unit multiplier and Observation status (and I think Series, especially if series_toggle: true)

Since this is the data repo, it can't rely on the series_toggle from the site repo. But I've added a config option called `indicator_options`, which can have a `non_disaggregation_columns` property. You can see the documentation of it here: https://github.com/open-sdg/sdg-build/pull/158/files#diff-de374cc323e2986aca98dd468f5c6813R94

So the idea is that if you don't use that config option, the defaults will kick in, which are: Year, Units, Series, Value, GeoCode, Observation status, Unit multiplier, and Unit measure. If, for example, you actually want Series to be a disaggregation, then you'd have to use this config option and leave out Series.

> Group is also showing up which is a disaggregation in the placeholder data files, so think we need to ensure only reported statistical indicators feed into report

This should be resolved now too.

NOTE: I should mention that this latest update touched a LOT of other files, so this PR is no longer just affecting the automated documentation. As such, we should really test out all the various inputs and outputs to make sure they still work. From my side, I've been testing with the Kazakhstan data, which uses the usual CSV data input, YAML metadata input, Open SDG output, and GeoJSON output.

LucyGwilliamAdmin:
@brockfanning ok I've just started doing a test with Kyrgyzstan data repo so SDMX-ML input:
https://lucygwilliamadmin.github.io/open-sdg-data-starter-1/disaggregations/

The main issue I've noticed so far is the translations in the By indicator section - some are showing concept.[CONCEPT] instead of the translation value

brockfanning:
That issue should be fixed now.

brockfanning:
@LucyGwilliamAdmin Sorry for the continued edits - I've seen some opportunity for refactoring, and also at least one bug. I'll let you know when it's ready again.

LucyGwilliamAdmin:
@brockfanning no problem at all!

When we download a csv from the disaggregation report and open it with Excel, completely numeric indicator codes (so 1-1-1, but not 3-c-1) are automatically getting changed to dates - is there anything that can be done in the code to stop this?

Re. order of sections within a disaggregation, is it worth either making the sections collapsible or having a section at top which lists (and links to) the different sections so user can easily see there's other sections and can get to them easily

brockfanning:
@LucyGwilliamAdmin I've just added anchor links to the top.

Regarding the indicator/date problem, there are a multiple options:
1. We can change the indicator numbers to something like `="1.1.1"` instead of just `1.1.1`, which will signal Excel that they need to be strings. This will work great in Excel, but has the downside that the CSV will contain this weird syntax which is really only for Excel.
2. We can change the indicator numbers to `Indicator 1.1.1` instead of `1.1.1`. This may clutter up the report, but should work fine. Or substitute any other textualizing of the number. Eg: `[1.1.1]`, `ID 1.1.1`, `#1.1.1`, `Ind 1.1.1`, etc.
3. We can create a second download button that is an actual Excel file (instead of a CSV) and has the column specifically flagged as a text column. This would involve more work than the other options.

Any thoughts?

AnnCorp:
@brockfanning My thoughts are to go for one of the formats in option (2) - as some of the indicator references  are treated as text anyway, may as well consistently make all the number-only indicator IDs behave like text too. What do you think @LucyGwilliamAdmin ?

LucyGwilliamAdmin:
@AnnCorp @brockfanning I agree - think option 2 would be best. I also agree `Indicator 1.1.1` may clutter so think one of the other  options e.g. `[1.1.1]`, `ID 1.1.1`, `#1.1.1`, `Ind 1.1.1` may be better

LucyGwilliamAdmin:
@brockfanning quick query:

What are the values in the `Number of instances` column? e.g. are there 3 indicators with `1 to 4` as an Age breakdown or does this mean something else?
image

Also, I've just noticed that the button in the top-left normally goes back to [Overview page](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/disagg-report/) except when clicking it from a disaggregation report page - then it goes back to Disaggregation report home page - is this right?

image

brockfanning:
@LucyGwilliamAdmin @AnnCorp I've added '#' in front of the indicator numbers (it seemed the least obtrusive). I've also removed the 'disaggregations' subfolder, and am now putting everything into the root folder, which should fix the problem with the header not linking back to the Overview page.

For "Number of instances" on the disaggregation detail pages, my intention was that it would be the number of times the value is used in a set of years. Or in other words, for each unique combination of disaggregations. So for example, take a disaggregation "Foo" with a value "Bar". Given this CSV data:

Year | Foo | Sex | Value
--- | --- | --- | ---
2018 | Bar | | 1.3
2019 | Bar | | 1.4
2018 | Bar | Female | 1.5
2019 | Bar | Female | 1.6

"Bar" would have 2 "instances" - one when it was used by itself, and one when it was used with "Female".

Hopefully the report is calculating that as I intended? I'm also open to any changes if you think that could be better described, or if you think a different metric would be more useful.

LucyGwilliamAdmin:
Ok, fab

Part of me is wondering if it's worth having another level so maybe instead of having `Number of instances` here, could have `Number of indicators` and the values in the `Value` column e.g. `1 to 4` become links, which go down to the new lower level
image

Example of how new lower level could look:
Slide1
Slide2

Not sure what others thoughts are @AnnCorp @brockfanning ?

Also not sure if possible/how much extra work this would be @brockfanning ?

brockfanning:
All of the "plumbing" is already written, so I think the extra work is not an issue. So the question is whether it's an improvement. Personally I think that would be a nice improvement.

brockfanning:
@LucyGwilliamAdmin I've made that change (detail pages for each disaggregation value) so you can see it in action.

Do you have any ideas for a better name for "Number of instances"? It's hard not to be wordy. I don't think I intended for that number to be used for any precise calculations, but more just for a general gauge of how "popular" a value is. Maybe "Number of disaggregation combinations using this value"?

LucyGwilliamAdmin:
@brockfanning great thank you - I will take a look

Definitely think "Number of disaggregation combinations using this value" describes it better than "Number of instances" - I'll have a think about other possible ways to word it though

brockfanning:
Guess we could shorten to "Disaggregation combinations using this value", as it would be pretty clear that it's a total.

AnnCorp:
@LucyGwilliamAdmin great suggestion for an improvement and @brockfanning  I think "Disaggregation combinations using this value" definitely clearer than 'Number of instances'.

brockfanning:
@LucyGwilliamAdmin I've just added a "Total rows" above each table. I'll stop now for feedback.

@AnnCorp To follow up on our discussion - an additional report could be an indication of progress towards complete disaggregation, according to some benchmark. If we had a CSV containing the required disaggregations for each indicator, we can generate percentages. We could have a report per-indicator, showing the progress percentage for each indicator. And it could also generate an overall percentage. We had talked about automating 17.18.1, but I realized that Open SDG does time-series charts, and an automated solution would only be able to see the current data. But definitely, each year, a country could take their overall percentage from the report and add it to their CSV file for 17.18.1.

Regardless of the solution there, I think that feature could be added in a separate PR. It could be an addition to what we have here, and wouldn't require any re-writing of anything.

AnnCorp:
@brockfanning absolutely agree - definitely a separate issue/PR.  Re 17.18.1, I think there's lots of potential ways to address it - lets discuss in the other ticket. Very exciting to be progressing all this with such pace!

LucyGwilliamAdmin:
@brockfanning Disaggregation report looks really good

Think I just have one final thing to add - Non reported indicators are still showing "Group" from placeholder data:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/disagg-report/disaggregations.html

brockfanning:
@LucyGwilliamAdmin Gotcha - before I was only checking that the indicator was statistical. I've just added a check to ensure it is also complete. Can you see if this helps?

LucyGwilliamAdmin:
Yes, this has worked :)

brockfanning:
@LucyGwilliamAdmin Great! I don't believe there are any pending issues - any objections to merging?

PR #165, opened 2020-08-03
Handle Obs elements without ObsValue

### LINKED ISSUE #164: SDMX input - support Obs without ObsValue
Right now, the SDMX input breaks if an `Obs` element does not contain an `ObsValue` element. This is because of this line: https://github.com/open-sdg/sdg-build/blob/master/sdg/inputs/InputSdmxMl_Structure.py#L125

Let's adjust that line so that it skips the loop if there is no ObsValue element.

brockfanning:
Fixes #164

### DIFF
--- a/sdg/inputs/InputSdmx.py
+++ b/sdg/inputs/InputSdmx.py
@@ -393,6 +393,8 @@ class InputSdmx(InputBase):
 
         # Create the Indicator objects.
         for indicator_id in indicator_data:
+            if not indicator_data[indicator_id]:
+                continue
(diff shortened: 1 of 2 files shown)


brockfanning:
@LucyGwilliamAdmin @jwestw This adds some safety code to catch some edge-cases in which certain elements/data is missing from the SDMX input. It's not too easy to test, however as evidence see how it is allowing these Cambodia proof-of-concept data to successfully build again: https://github.com/brockfanning/sdg-data-cambodia/actions

PR #167, opened 2020-08-12
Service to copy indicator data files to the build for direct downloads

brockfanning:


LucyGwilliamAdmin:
@brockfanning I was wrong actually - still seem to be having an issue with translation key:

https://github.com/LucyGwilliamAdmin/open-sdg-data-starter-1/runs/1253181846?check_suite_focus=true

PR #168, opened 2020-08-13
Fix CSV metadata input

LucyGwilliamAdmin:


LucyGwilliamAdmin:
Naming was inconsistent so was getting an error, when trying to use CSV metadata input.

brockfanning (dismissed):

PR #172, opened 2020-09-27
Output untranslated build along with translation builds

brockfanning:
This adds an untranslated build along with the translated builds. Whereas previously the output of translated builds would include a subfolder for each language, now there is an additional subfolder called "untranslated".

This should not change any current behavior/functionality, but it allows for the usage of this untranslated data/metadata. There may be other future uses for this, but at current the first intended use is in an indicator configuration form, which needs to be populated with the untranslated metadata.

### DIFF
--- a/sdg/open_sdg.py
+++ b/sdg/open_sdg.py
@@ -117,6 +117,8 @@ def open_sdg_build(src_dir='', site_dir='_site', schema_file='_prose.yml',
         if options['languages']:
             # If languages were provide, perform a translated build.
             status = status & output.execute_per_language(options['languages'])
+            # Also provide an untranslated build.
+            status = status & output.execute('untranslated')
         else:
             # Otherwise perform an untranslated build.
             status = status & output.execute()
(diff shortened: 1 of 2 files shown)


brockfanning:
One issue with this: 'untranslated' is getting added to the list of languages in Open SDG.

brockfanning:
@LucyGwilliamAdmin @jwestw This one would help with the site config forms, so I was wondering if you would have time to take a look. The main thing I think is to ensure that the data builds still work as normal. If possible it would be good to test with both a mono-lingual and multi-lingual site.

LucyGwilliamAdmin:
Here's for a single language site:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/output-untranslated-build/untranslated/

Looks good to me

LucyGwilliamAdmin:
Multi-lingual:
https://lucygwilliamadmin.github.io/open-sdg-data-starter-1/untranslated/data/1-b-1-1.json
https://lucygwilliamadmin.github.io/open-sdg-data-starter-1/en/data/1-b-1-1.json
https://lucygwilliamadmin.github.io/open-sdg-data-starter-1/ru/data/1-b-1-1.json

Multilingual also looks good to me

LucyGwilliamAdmin (approved):

PR #173, opened 2020-10-18
Use column as a default group when translating data

brockfanning:
Right now, when translating data, `data` is always used as the default group. For example, if a cell contains `foo` then it will be translated as if it were a "translation key" of `data.foo`. This PR first tries to use the column name as the default group, while still falling back to `data`. For example, if a cell in the `baz` column contains `foo`, then it will be translated as if were `baz.foo`, or if that doesn't work, `data.foo`.

### DIFF
--- a/sdg/Indicator.py
+++ b/sdg/Indicator.py
@@ -254,8 +254,8 @@ class Indicator:
                 return {key: translate_meta(value) for (key, value) in text.items()}
             # Otherwise treat as a string.
             return translation_helper.translate(text, language, default_group='data')
-        def translate_data(text):
-            return translation_helper.translate(text, language, default_group='data')
+        def translate_data(text, column):
+            return translation_helper.translate(text, language, default_group=[column, 'data'])
         def translate_data_columns(text):
             special_columns = ['Year', 'Value', 'Units']
             if text in special_columns:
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin:
I think this may be a side note but when testing this with Kyrgyzstan sandpit, I'm getting this error:
https://github.com/LucyGwilliamAdmin/open-sdg-data-starter-1/actions/runs/332142413

```
File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sdg/OutputDocumentationService.py", line 242, in write_disaggregation_value_detail_page
    html = self.get_html(disaggregation + ': ' + disaggregation_value, service.get_disaggregation_value_detail_template().format(
TypeError: can only concatenate str (not "float") to str
```

I'm wondering if they maybe have a disaggregation named using a number instead of a string?

LucyGwilliamAdmin:
@brockfanning just tried this out:

https://lucygwilliamadmin.github.io/open-sdg-site-starter-1/en/1-4-1/
For 1.4.1 I changed one of the columns to baz, and one of the rows in that column to foo.
I then created a new translations file called baz.csv which contains the translations for foo.

https://lucygwilliamadmin.github.io/open-sdg-site-starter-1/en/10-b-1/
For 10-b-1 I changed just one row in one column to foo (didn't change column name).
I then added a translation key to data.csv for foo.

Both are appearing as I expected, but I have done the test right?

brockfanning:
> 
> I'm wondering if they maybe have a disaggregation named using a number instead of a string?

@LucyGwilliamAdmin Very possibly - there's a ticket about that still (https://github.com/open-sdg/open-sdg/issues/327).

Your tests both sound good to me.

LucyGwilliamAdmin:
Great - will approve and try out the untranslated build next.

I deleted my comment as found the issues after - somehow I ended up with 2 white spaces before the Year column in one of the CSVs

LucyGwilliamAdmin (approved):

PR #174, opened 2020-10-20
Disaggregation stats in Open SDG output

brockfanning:
This adds an "endpoint" at /stats/disaggregation.json to the Open SDG output. This includes the following data:

* disaggregated: Number of indicators that are fully disaggregated
* not_disaggregated: Number of indicators that are not fully disaggregated
* total: Number of statistical indicators (this will be equal to the total of "disaggregated" and "not_disaggregated")
* disaggregated_percent: Percentage of indicators that are fully disaggregated (percentage of "total")
* not_disaggregated_percent: Percentage of indicators that are not fully disaggregated (percentage of "total")

The way that it determines "fully disaggregated" is dependent on a metadata field called `expected_disaggregations`, which should be a list of column names.

For example, if an indicator is expected to be disaggregated by "Age" and "Sex", then this metadata field should be something like (in YAML):

```
expected_disaggregations:
  - Age
  - Sex
```

If the indicator's data itself has both an `Age` and a `Sex` column, then it is considered "fully disaggregated".

If an indicator does not have an `expected_disaggregations` metadata field, then it will automatically be considered fully disaggregated.

The next step is to add code to Open SDG to use this data.

### DIFF
--- a/sdg/DisaggregationStatusService.py
+++ b/sdg/DisaggregationStatusService.py
@@ -77,7 +77,7 @@ class DisaggregationStatusService:
         status = {
             'statuses': [{
                 'value': 'fully_dissaggregated',
-                'translation_key': 'status.fully_disaggregated',
+                'translation_key': 'Fully disaggregated',
             }],
             'overall': {
                 'statuses': [],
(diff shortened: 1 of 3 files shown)


LucyGwilliamAdmin:
I started testing this and then realised it's still marked as draft, but leaving a comment here about my error:
https://github.com/ONSdigital/sdg-data/runs/1365907679?check_suite_focus=true

PR #177, opened 2020-11-22
Sdmx output "light" version

brockfanning:
This is an SDMX output that depends on a few assumptions:

1. There is a valid DSD available
2. The disaggregation columns in the data already correspond to dimensions in that DSD.
3. The disaggregation values in those columns already correspond to codelist codes for that dimension in that DSD.

## Usage

This output can be enabled with the Open SDG YAML-based config (data_config.yml, etc.). To use it in Open SDG, you would add this to the data config:

```
sdmx_output:
    dsd: my-dsd-file.xml
```

The DSD can also be remote:

```
sdmx_output:
    dsd: https://example.com/my-dsd-file
```

## SDMX Input and Translation update

When we originally added the ability for inputting SDMX data and SDMX translations, we coded them to be input as translation keys - like "codelist.my-codelist" for disaggregation column names and "my-codelist.my-code" for disaggregation values. However, in doing this PR, I realized we should have used dimension IDs and not codelist IDs, for the column names. I also realized that it was not necessary to input them as translation keys, and that they could just be IDs.

So this PR also updates that: changes the way that SDMX data is input, that way the SDMX translations are input, and the way that indicator data is translated. This shift should all be invisible, and SDMX input should continue to behave the same as before. It's just that the internal stuff is different.

## Third-party SDMX library

On the tech side, this is our first attempt at using a third-party SDMX library. In this case, I used: https://github.com/khaeru/sdmx

## Python version

A consequence of using the SDMX library mentioned above is that it would make sdg-build require Python 3.7 or higher. In other words Python 3.5 and 3.6 would no longer work. You can see this by the test failures on the PR. I don't have any particular problem with this, but it's worth discussing.


brockfanning:
@LucyGwilliamAdmin @jwestw I believe this is ready for testing. I think that Kyrgyzstan is the only implementation that's in a good spot to test this. (Even though technically they don't need this, because they already have SDMX files.)

LucyGwilliamAdmin:
Looks good:
https://lucygwilliamadmin.github.io/open-sdg-data-starter-1/sdmx-output.html

Should I try running a file through a validation tool?

brockfanning:
@LucyGwilliamAdmin That sounds great!

LucyGwilliamAdmin:
@brockfanning just ran through validation and the file that was produced by platform had quite a lot of errors so took a look at the file and it seems to be missing quite a lot of info.

File produced by platform (https://lucygwilliamadmin.github.io/open-sdg-data-starter-1/sdmx/1-1-1-1a.xml):
image

File that was uploaded (https://github.com/LucyGwilliamAdmin/open-sdg-data-starter-1/blob/develop/data/1.1.1.1a.xml):
image

brockfanning:
Very helpful, thanks @LucyGwilliamAdmin.

brockfanning:
@LucyGwilliamAdmin I'm thinking that it is actually the SDMX input that's dropping some disaggregations. You can get more disaggregations by setting `drop_singleton_dimensions` to false. For example:

```
inputs:
  - class: InputSdmxMl_Multiple
    path_pattern: data/*.xml
    dsd: 'SDG_DSD.KG.xml'
    import_codes: true
    drop_singleton_dimensions: false
```

Even with that though, any disaggregation will a value of "_T" will be dropped automatically, because of [this code](https://github.com/open-sdg/sdg-build/blob/master/sdg/inputs/InputSdmx.py#L327-L328). If I set drop_singleton_dimensions to false and also comment out those lines dropping the _T's, here is an example of what I get:

```
          <gen:Obs>
            <gen:ObsKey>
                <gen:Value id="FREQ" value="A" />
                <gen:Value id="REPORTING_TYPE" value="N" />
                <gen:Value id="SERIES" value="_L_KG_010101_1A" />
                <gen:Value id="REF_AREA" value="KG" />
                <gen:Value id="SEX" value="_T" />
                <gen:Value id="AGE" value="_T" />
                <gen:Value id="URBANISATION" value="_T" />
                <gen:Value id="INCOME_WEALTH_QUANTILE" value="_T" />
                <gen:Value id="EDUCATION_LEV" value="_T" />
                <gen:Value id="OCCUPATION" value="_T" />
                <gen:Value id="CUST_BREAKDOWN" value="_T" />
                <gen:Value id="COMPOSITE_BREAKDOWN" value="_T" />
                <gen:Value id="DISABILITY_STATUS" value="_T" />
                <gen:Value id="ACTIVITY" value="_T" />
                <gen:Value id="PRODUCT" value="_T" />
            </gen:ObsKey>
            <gen:ObsValue value="5.3" />
            <gen:Attributes>
                <gen:Value id="UNIT_MEASURE" value="PT" />
                <gen:Value id="TIME_DETAIL" value="2010" />
                <gen:Value id="COMMENT_TS" value="" />
            </gen:Attributes>
        </gen:Obs>
```

This corresponds to this source:

```
<Series
  FREQ="A"
  REPORTING_TYPE="N"
  SERIES="_L_KG_010101_1A"
  REF_AREA="KG"
  SEX="_T"
  AGE="_T"
  URBANISATION="_T"
  INCOME_WEALTH_QUANTILE="_T"
  EDUCATION_LEV="_T"
  OCCUPATION="_T"
  CUST_BREAKDOWN="_T"
  COMPOSITE_BREAKDOWN="_T"
  DISABILITY_STATUS="_T"
  ACTIVITY="_T"
  PRODUCT="_T"
  UNIT_MEASURE="PT">
    <Obs
      TIME_PERIOD="2010"
      OBS_VALUE="5.3"
      OBS_STATUS="A"
      UNIT_MULT="0"
      NATURE="C"
      TIME_DETAIL="2010"
>
</Obs>
```

So it gets most of the attributes, but does still lack these:
* TIME_PERIOD
* OBS_STATUS
* UNIT_MULT
* NATURE

LucyGwilliamAdmin:
@brockfanning ok, I just tried the same and file now looks like this so pretty much what you said would happen - I did need to remove `REF_AREA|KG: ''` in order for `<gen:Value id="REF_AREA" value="KG"/>` to show instead of `<gen:Value id="REF_AREA" value=""/>`
https://lucygwilliamadmin.github.io/open-sdg-data-starter-1/sdmx/1-1-1-1a.xml

This does of a major effect on platform:
https://lucygwilliamadmin.github.io/open-sdg-site-starter-1/en/1-1-1-1a/

Although a change I made in the build alter_data function didn't seem to affect the SDMX output

brockfanning:
@LucyGwilliamAdmin I'd like to take a look at the data_alter issue - can you link me to that function?

I notice that of the four things I mentioned were lacking, three of them (OBS_STATUS, UNIT_MULT, and NATURE) are all "attributes". I wonder if the SDMX input is not importing attributes.

Another question I'm wondering about is TIME_DETAIL vs TIME_PERIOD. I arbitrarily chose TIME_DETAIL as what this output maps the "Year" column to. But should I maybe have chosen TIME_PERIOD? Or both?

Regarding the issues with the Kyrgyzstan site if you change the input - that's a tricky question. My first reaction is that really you wouldn't want to use this with that site, since you already have SDMX there. I only suggested trying this out with Kyrgyzstan because that is the only site I know of where we can be sure that the data conforms to a DSD, which is the main requirement of using this output.

In practice I am imagining the main use-case for this output is a country that has their data conforming to a DSD, but is still maintaining the data in CSV files.

LucyGwilliamAdmin:
@brockfanning I don't think the data_alter is an issue but it was just an observation that changing things in the dimension map was affecting the SDMX output whilst making changes using the data_alter function doesn't

Yes, you're right - I think I forgot about that - Maybe I could try replacing an SDMX-ML file with a CSV which meets the DSD standards and then seeing what the SDMX output looks like?

brockfanning:
@LucyGwilliamAdmin That sounds great!

LucyGwilliamAdmin:
@brockfanning ok - I'll give it a go. I do have some questions first though?

- Should all terms be DSD codes?
- If a disaggregation isn't present does a column need to be present?
- Does _T need to be where blank cells would normally be?

LucyGwilliamAdmin:
I'm also thinking about those concepts that weren't being added - we added UNIT_MULT and OBS_STATUS to our CSVs in UK in preparation for SDMX as they are required concepts so I'm wondering if that's got something to do with it - some sort of value needs to be explicitly set per observation.

brockfanning:
@LucyGwilliamAdmin Good questions - here are my thoughts:
1. Column names should be the "id" of the Dimension element in the DSD. Column values should be the "id" of the Code element in the DSD
2. It can be left out, and I believe it will also be left out of the SDMX output. This is a question for me too - does SDMX output need to have all disaggregations for all observations - even if they are empty?
3. Good question - I think for ease of use we want to allow the user to just have blank cells. But it is still an open question for me - should the output change those blank cells to "_T" or leave them blank.

So I don't really have answers for 2 and 3, unfortunately. At the moment I believe that missing disaggregations will be left out of the output, and blank cells will appear as empty strings or left out completely (not sure which).

LucyGwilliamAdmin:
@brockfanning thank you -re. 1 does this include special columns e.g. Year, Value, Units, Series?

brockfanning:
@LucyGwilliamAdmin I believe those special columns can continue to be "Year", "Value", etc.

LucyGwilliamAdmin:
@brockfanning ok - so this is the CSV I created:
https://github.com/LucyGwilliamAdmin/open-sdg-data-starter-1/blob/develop/data/1-1-1-1a.csv

I kept all columns which had some difference in rows

This is the SDMX output:
https://lucygwilliamadmin.github.io/open-sdg-data-starter-1/sdmx/1-1-1-1a.xml

I have ran it through SDMX Converter validation and still lots of errors which I'm going to try to make sense of.

One thing I picked up along the way which I can see is an issue is:

- The code IDs for UNIT_MULT are numerical (exponent in base 10) e.g. 0 (Units), 1 (Tens), 2 (Hundreds), 3 (Thousands), etc...

LucyGwilliamAdmin:
Another thing I'm thinking about is time series level concepts. I'm wondering if maybe they would be set in indicator config file? e.g. FREQ, REPORTING_TYPE, NATURE

LucyGwilliamAdmin:
I think quite a few errors are being caused by the header - think it needs to look more like this:
image
than like this:
image

I think we've discussed before for diff reasons but wondering if there's any impact of having generic SDMX instead of structure specific?

LucyGwilliamAdmin:
I'm thinking this error might mean that re. 2 above SDMX output does need to have all disaggregations for all observations, even if they are all empty

image

brockfanning:
Great, thanks @LucyGwilliamAdmin I'll look at these and make some adjustments.

brockfanning:
@LucyGwilliamAdmin I just pushed a commit that will add empty strings for any dimensions/attributes that are missing. I'm not sure whether this will work, but just thought I'd try it. It's not clear to me how we would decide on a valid value to set for attributes/dimensions that are not in the CSV. Maybe we could screenshare sometime so I can learn how to do this validation? In the meantime does this reduce the number of validation errors?

About UNIT_MULT, do you mean that we would need to resolve [this numeric disaggregation issue](https://github.com/open-sdg/open-sdg/issues/327) before UNIT_MULT could be put into the CSV?

About Generic vs StructureSpecific, to be honest I don't know how I would control that, using this third-party SDMX library. I guess whatever it's outputting now is the default, or maybe the only option. I can investigate more though.

brockfanning:
@LucyGwilliamAdmin Actually hold off on testing - I see locally that the empty strings prevent the output from building.

brockfanning:
@LucyGwilliamAdmin Ok I think the issue with empty strings is fixed. Now it will default to "_T" for all missing dimensions, and it will default to "" for all missing attributes. Also you can now pass in your own default values like this for example:

```
sdmx_output:
    dsd: my-dsd-file.xml
    default_values:
        AGE: somedefaultvalue
        NATURE: somedefaultvalue
```

I imagine that this might be necessary if "_T" is not appropriate for some dimensions.

brockfanning:
@LucyGwilliamAdmin I just added another way to control the default values of attributes/dimensions: by including them in the indicator's metadata. For example, if the indicator's metadata has a field called "NATURE" then that will be used as a default value for "NATURE" for every row, in the case that it is not in the CSV file.

LucyGwilliamAdmin:
@brockfanning just coming back to this but all of a sudden getting an xlrd library error:
https://github.com/LucyGwilliamAdmin/open-sdg-data-starter-1/runs/1559662736?check_suite_focus=true#step:5:33

LucyGwilliamAdmin:
@brockfanning looks like xlrd has removed support for everything but .xls:
https://xlrd.readthedocs.io/en/latest/changes.html#id1

brockfanning:
@LucyGwilliamAdmin Interesting - I don't think that's related to this PR but maybe it deserves a separate ticket. The only code we have related to Excel is [InputExcelMeta](https://github.com/open-sdg/sdg-build/blob/master/sdg/inputs/InputExcelMeta.py), and that only uses Pandas' "pd.ExcelFile". But I guess pd.ExcelFile relies on having something like xlrd installed? Would it help to add "xlrd" to your requirements.txt file?

LucyGwilliamAdmin:
@brockfanning yeah - that's what it looks like. I did have a load of libraries listed in requirements.txt but then seen that they weren't in site-starter so removed them - I'll add xlrd back in now though to see if helps

LucyGwilliamAdmin:
I've just added xlrd to my requirements and I'm now getting this error:
https://github.com/LucyGwilliamAdmin/open-sdg-data-starter-1/runs/1563351721?check_suite_focus=true#step:5:37

I'll disable the metadata Excel input for now so I can continue with this

LucyGwilliamAdmin:
@brockfanning disabling it worked and I've been able to rerun - new output is here:
https://lucygwilliamadmin.github.io/open-sdg-data-starter-1/sdmx/1-1-1-1a.xml

I've set NATURE and UNIT_MULT in indicator metadata.

I've run through validation and got the following results:

- `FREQ` by default is `_T` but `_T` isn't valid in `FREQ` codelist
- `REPORTING_TYPE` by default is `_T` but `_T` isn't valid in `REPORTING_TYPE` codelist

I think the following attributes aren't required so could maybe left out of SDMX output unless set as when not set by default they are "", which is causing error in validation

```
<gen:Value id="COMMENT_OBS" value=""/>
<gen:Value id="TIME_COVERAGE" value=""/>
<gen:Value id="UPPER_BOUND" value=""/>
<gen:Value id="LOWER_BOUND" value=""/>
<gen:Value id="BASE_PER" value=""/>
<gen:Value id="SOURCE_DETAIL" value=""/>
<gen:Value id="COMMENT_TS" value=""/>
<gen:Value id="GEO_INFO_URL" value=""/>
<gen:Value id="GEO_INFO_TYPE" value=""/>
<gen:Value id="CUST_BREAKDOWN_LB" value=""/>
<gen:Value id="DATA_LAST_UPDATE" value=""/>
```

brockfanning:
@LucyGwilliamAdmin Thanks for helping work through these. I've made those changes.

LucyGwilliamAdmin:
@brockfanning just reran:

- TIME_PERIOD is an attribute in structure specific files and when I convert Kyrgyzstan's actual file to generic, TIME_PERIOD doesn't exist

Current output is:

```
<gen:ObsValue value="5.3"/>
<gen:Attributes>
    <gen:Value id="OBS_STATUS" value="A"/>
    <gen:Value id="UNIT_MULT" value="0"/>
    <gen:Value id="UNIT_MEASURE" value="PT"/>
    <gen:Value id="NATURE" value="C"/>
    <gen:Value id="TIME_DETAIL" value="2010"/>
</gen:Attributes>
```

Converted original file looks like this:

```
-<generic:Obs>
    <generic:Time>2010</generic:Time>
    <generic:ObsValue value="5.3"/>
   -<generic:Attributes>
        <generic:Value value="A" concept="OBS_STATUS"/>
        <generic:Value value="C" concept="NATURE"/>
        <generic:Value value="0" concept="UNIT_MULT"/>
        <generic:Value value="2010" concept="TIME_DETAIL"/>
    </generic:Attributes>
</generic:Obs>
```



Also I've just noticed that output dimensions section is called ObsKey but converted file's SeriesKey

```
<gen:Obs>
    <gen:ObsKey>
        <gen:Value id="FREQ" value="A"/>
        <gen:Value id="REPORTING_TYPE" value="N"/>
        <gen:Value id="SERIES" value="_L_KG_010101_1A"/>
        <gen:Value id="REF_AREA" value="KG"/>
        <gen:Value id="SEX" value="_T"/>
        <gen:Value id="AGE" value="_T"/>
        <gen:Value id="URBANISATION" value="_T"/>
        <gen:Value id="INCOME_WEALTH_QUANTILE" value="_T"/>
        <gen:Value id="EDUCATION_LEV" value="_T"/>
        <gen:Value id="OCCUPATION" value="_T"/>
        <gen:Value id="CUST_BREAKDOWN" value="_T"/>
        <gen:Value id="COMPOSITE_BREAKDOWN" value="_T"/>
        <gen:Value id="DISABILITY_STATUS" value="_T"/>
        <gen:Value id="ACTIVITY" value="_T"/>
        <gen:Value id="PRODUCT" value="_T"/>
        <gen:Value id="TIME_PERIOD" value="_T"/>
</gen:ObsKey>
```

Converted original file:

```
-<generic:SeriesKey>
    <generic:Value id="FREQ" value="A"/>
   <generic:Value id="REPORTING_TYPE" value="N"/>
   <generic:Value id="SERIES" value="_L_KG_010101_1A"/>
   <generic:Value id="REF_AREA" value="KG"/>
   <generic:Value id="SEX" value="_T"/>
   <generic:Value id="AGE" value="_T"/>
   <generic:Value id="URBANISATION" value="_T"/>
   <generic:Value id="INCOME_WEALTH_QUANTILE" value="_T"/>
   <generic:Value id="EDUCATION_LEV" value="_T"/>
   <generic:Value id="OCCUPATION" value="_T"/>
   <generic:Value id="CUST_BREAKDOWN" value="_T"/>
   <generic:Value id="COMPOSITE_BREAKDOWN" value="_T"/>
   <generic:Value id="DISABILITY_STATUS" value="_T"/>
   <generic:Value id="ACTIVITY" value="_T"/>
   <generic:Value id="PRODUCT" value="_T"/>
</generic:SeriesKey>
```

brockfanning:
@LucyGwilliamAdmin I'm also comparing to the [output of the .Stat suite that Cambodia is running](https://nsiws-stable-camstat-live.officialstatistics.org/rest/data/KH_NIS,DF_SDG_KH,1.2/A..............?startPeriod=2008&endPeriod=2019). In that output I see TIME_PERIOD as a dimension, that's always set to the year. So I've just pushed a commit that always sets TIME_PERIOD to the year.

I can't explain those differences from the converted StructureSpecific but I'll do some research.

LucyGwilliamAdmin:
@brockfanning ah that's fine if that's what in the Camstat one - I'll refer to that aswell for structure

LucyGwilliamAdmin:
@brockfanning it does look like CamStat also have their dimensions within a `SeriesKey` element rather than an `ObsKey` element

LucyGwilliamAdmin:
It does look like we don't have many errors left in validation now 👍 

4 unique ones:

- `Error While Processing XML : Expected element 'ID@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' instead of 'Test@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' here in element Header@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message`
- `Error While Processing XML : Expected element 'ID@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' instead of 'Structure@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' here in element Header@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message`
- `Error While Processing XML : Expected element 'ID@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' before the end of the content in element Header@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message`

I think the above 3 relate to the message header

- `Non flat dataSet expected, (dimension At Observation is not 'All Dimensions') but flat dataset encountered. Flat dataSets contain only Obs elements` (reoccurs approx 105 times)

I think this might actually be related to issue mentioned above (use of `ObsKey` instead of `SeriesKey`

brockfanning:
Ok, cool. Yes I'm also seeing a difference in "flat" vs "time series". Here is the Cambodia SDMX as "flat", which seems closer to what this output is creating: https://nsiws-stable-camstat-live.officialstatistics.org/rest/data/KH_NIS,DF_SDG_KH,1.2/A..............?startPeriod=2008&endPeriod=2019&dimensionAtObservation=AllDimensions

I'll see if it's possible to create time series SDMX with this library.

LucyGwilliamAdmin:
Alternatively, looking at Cambodia flat file, they have `<message:Structure structureID="KH_NIS_DF_SDG_KH_1_2" dimensionAtObservation="AllDimensions">`

image

The output doesn't currently have that, but error does mention '(dimension At Observation is not 'All Dimensions')' so if not possible to change output type, maybe there's a way to set that which might resolve the error (not sure though)

brockfanning:
@LucyGwilliamAdmin I haven't tackled the header issues yet, but I've just pushed an attempt to switch over to time series datasets.

brockfanning:
@LucyGwilliamAdmin Ok, got the header stuff in there now, so this is ready for another test.

LucyGwilliamAdmin:
@brockfanning just trying to test this again but getting an error:
https://github.com/LucyGwilliamAdmin/open-sdg-data-starter-1/runs/1680666045?check_suite_focus=true

brockfanning:
Thanks @LucyGwilliamAdmin I'm investigating now.

brockfanning:
@LucyGwilliamAdmin I just pushed a commit that I believe fixes that error.

LucyGwilliamAdmin:
@brockfanning I've just tried again and not getting that error anymore - getting this now though:
https://github.com/LucyGwilliamAdmin/open-sdg-data-starter-1/runs/1682160467?check_suite_focus=true#step:5:1271

brockfanning:
@LucyGwilliamAdmin That should now be fixed too.

LucyGwilliamAdmin:
Great thanks - SDMX outputting now. From validation:

- Invalid Date Format `2021-01-11T16:16:54.503121` - I think this needs to be `2021-01-11T16:16:54`

- The dataset contains an observation with a missing a mandatory attribute : OBS_STATUS - OBS_STATUS is in CSV so not sure if being carried through to output?

- The dataset contains an observation with a missing a mandatory attribute : UNIT_MULT - same as above but UNIT_MULT set in indicator config file

- The dataset contains an observation with an attribute `UNIT_MEASURE` that is not defined in the DSD **_&_** The dataset series is missing a mandatory attribute : UNIT_MEASURE - not sure if UNIT_MEASURE is being applied at correct level. In SDMX generic file I converted from Kyrgyzstan structure specific file the UNIT_MEASURE attribute is in a different place:

Converted Kyrgyzstan file:
image

Open SDG output:
image

brockfanning:
Ok, cool. I will take another pass at this after we've got 1.2.0 wrapped up.

LucyGwilliamAdmin:
Also a couple more re. the header:

- Error While Processing XML : Expected element 'ID@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' instead of 'Test@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' here in element Header@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message
- Error While Processing XML : string value 'my-id-1610381814.503121' does not match pattern for IDType in namespace http://www.sdmx.org/resources/sdmxml/schemas/v2_1/common - from what I can see, these always seem to start with IREF
- Error While Processing XML : Expected element 'Test@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' instead of 'Prepared@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' here in element Header@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message
- Error While Processing XML : Expected element 'Test@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' instead of 'Sender@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' here in element Header@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message
- Error While Processing XML : Expected element 'Test@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' instead of 'Structure@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' here in element Header@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message
- Error While Processing XML : Expected element 'Test@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message' before the end of the content in element Header@http://www.sdmx.org/resources/sdmxml/schemas/v2_1/message

I've kept these separate as I will try to find out what they mean

brockfanning:
@LucyGwilliamAdmin I think those last errors are related to the order of the elements in the "mes:Header" element. It seems like the validator expects the first element to be "mes:ID". And since it is not, it causes a chain reaction of other similar errors. I would not have expected that the order of XML elements was important, but I'm asking in the the [sdmx library](https://github.com/khaeru/sdmx) to see if they can help. The order could be pretty easily tweaked in that library, if needed.

I believe I've fixed the other issues, so this is ready for another test.

Note: In order to pull in those observation-level attributes (like OBS_STATUS, etc.) I needed to change the SDMX input classes. So it would be good to test this with Kyrgyzstan, to make sure that the additional attributes don't cause a problem (separate from any testing of SDMX output).

LucyGwilliamAdmin:
@brockfanning I'll give it another test - I'll also manually swap the order to see if that resolves the error message

LucyGwilliamAdmin:
@brockfanning I'm also surprised that the order would make a difference but I've just swapped the ID and Test elements and only two errors remain

- Error While Processing XML : string value 'IREF1611243613.619202' does not match pattern for IDType in namespace http://www.sdmx.org/resources/sdmxml/schemas/v2_1/common

  I think this is caused by the decimal point in the ID. Also I'm not sure how unique that ID needs to be?

- Error While Processing XML : string value 'open-sdg/sdg-build@1.3.0' does not match pattern for IDType in namespace http://www.sdmx.org/resources/sdmxml/schemas/v2_1/common

  I think this is caused by punctuation in Sender but I imagine we would want some config option for this?

brockfanning:
I believe I've found the description of that attribute: https://github.com/sdmx-twg/sdmx-ml-v2_1/blob/master/schemas/SDMXCommonReferences.xsd#L33

Given that, I should be able to make it pass validation. But I'll also try to make them configurable.

brockfanning:
@LucyGwilliamAdmin I believe that all the header issues should be resolved now except for the order of the elements. The problem of the order of the elements has now been [fixed in the sdmx project](https://github.com/khaeru/sdmx/commit/16e9970be06cdd570857f54d06c20a0f33eb0307), but we will not see the change until they have their next release.

## Testing with Kyrgyzstan

Before merging this, I think we should confirm that the Kyrgyzstan site still works with the new code - regardless of any SDMX output. Because this PR changes the SDMX input, and Kyrgyzstan is already using it, I think this would make sense. I've tested it out locally, and I needed to make a few config changes. [Here](https://github.com/brockfanning/open-sdg-data-starter-1/commit/b492495df83007b062f8542c0e52b98a16daf3a8) is an example of what I needed to do. The changes were:

1. Change "drop_dimensions" to a list. This is actually not new, and I suspect that drop_dimensions was not having the intended effect before.
2. Add COMMENT_OBS to drop_dimensions. I noticed that it started appearing, and I didn't think it would be desired. (The SDMX input now finds more attributes that it was ignoring before.)
3. Change "import_translation_keys" to "import_codes". This is not actually required because it is backwards-compatible, but I suggest the change to avoid confusion.
4. Add the same "dimension_map" to the translation input. This is necessary because the translation input now expects the same column names as in the SDMX input. So if the SDMX input is renaming column names, the translation input needs to do the same.

One side-effect I noticed was that the "Series" disaggregation label is no longer translated. A quick fix for this would be if they could switch to using `series_toggle: true`. (This is gong to be our general recommendation going forward, and we've already added it to the site starter.) However if this is not possible, we'll have to investigate another solution.

LucyGwilliamAdmin:
@brockfanning great, ran back through validation and no errors (apart from the one that will be fixed by next library release) - does the sender configurable option need to be in this PR or what that be separate?

Looking at how this affects input now. Kyrgyzstan actually requested that they be able to have the series in dropdown rather than toggle so not sure they'll want to switch

LucyGwilliamAdmin:
@brockfanning when I try to set dimension map for translations I get an error:
https://github.com/sustainabledevelopment-lucy/open-sdg-data-starter/runs/1762000819?check_suite_focus=true#step:5:17

brockfanning:
@LucyGwilliamAdmin Ok, yeah that looks like it wasn't pointed at this branch.

Actually I realized that "Series" not being translated is already happening on the current site. And an easy fix for this is to remove SERIES from the dimension_map, like [here](https://github.com/brockfanning/open-sdg-data-starter-1/commit/acb31d9a91fe8cd4c47eb76305630699b9688364).

LucyGwilliamAdmin:
Yeah, removing Series from dimension map has fixed it. One reason for possibly not having series as disaggregation is the possibility of large differences in values:

https://sdg-kyrgyzstan.github.io/open-sdg-site-starter/en/1-3-1-1g/

brockfanning:
Make sense.

LucyGwilliamAdmin:
Checking this with Kyrgyzstan site now. One issue I've found so far is this - Age value isn't automatically being selected but only seems to be an issue with 4.3.1.1 so far

Pointing at this branch:
https://sustainabledevelopment-lucy.github.io/open-sdg-site-starter/en/4-3-1-1/

Pointing at 1.2.0-beta2:
https://sdg-kyrgyzstan.github.io/open-sdg-site-starter/en/4-3-1-1/

brockfanning:
@LucyGwilliamAdmin Ah, that's an important note - I guess this means that the values for "data_start_values" and similar indicator configuration will need to be changed. For this example, the existing configuration...

```
data_start_values:
  - field: concept.AGE
    value: CL_AGE.Y15T17
  - field: concept.AGE
    value: CL_AGE.Y17T20
  - field: concept.AGE
    value: CL_AGE.Y17T24
  - field: concept.EDUCATION_LEV
    value: CL_EDUCATION_LEV.AGG_3_4
  - field: concept.EDUCATION_LEV
    value: CL_EDUCATION_LEV.ISCED11_5
  - field: concept.EDUCATION_LEV
    value: CL_EDUCATION_LEV.ISCED11_6_7
```

Would need to be changed to this (I think):

```
data_start_values:
  - field: AGE
    value: Y15T17
  - field: AGE
    value: Y17T20
  - field: AGE
    value: Y17T24
  - field: EDUCATION_LEV
    value: AGG_3_4
  - field: EDUCATION_LEV
    value: ISCED11_5
  - field: EDUCATION_LEV
    value: ISCED11_6_7
```

Which is nice because it's simpler, but definitely a backwards compatibility issue. Though Kyrgyzstan is the only site that we know of that would be using this.

brockfanning:
@LucyGwilliamAdmin Actually I'm trying that out locally and not having success - let me investigate a bit more.

brockfanning:
@LucyGwilliamAdmin Ok, so it's not actually simpler now, just different... Here is the new config needed:

```
data_start_values:
  - field: AGE.AGE
    value: AGE.Y15T17
  - field: AGE.AGE
    value: AGE.Y17T20
  - field: AGE.AGE
    value: AGE.Y17T24
  - field: EDUCATION_LEV.EDUCATION_LEV
    value: EDUCATION_LEV.AGG_3_4
  - field: EDUCATION_LEV.EDUCATION_LEV
    value: EDUCATION_LEV.ISCED11_5
  - field: EDUCATION_LEV.EDUCATION_LEV
    value: EDUCATION_LEV.ISCED11_6_7
```

I'm not really happy with that - since there is no way that users would know to do that. I'm going to think on possible ways to make it simpler.

LucyGwilliamAdmin:
Ah ok - yes can see how that could be quite difficult to set.

Other than that, Kyrgyzstan site looks good! Do we need to do any other testing for output? Maybe a more complex file? Or even some diff files?

brockfanning:
I don't have any other ideas for testing.

Regarding those translation-key issues - I think I might be able to improve it, but it would be done elsewhere, in jekyll-open-sdg-plugins. It should be fine to do this later, after merging this, as a future enhancement.

LucyGwilliamAdmin:
Ok - no probs. We (UK) are going to use this to convert some of our more simpler global files so that could be useful for testing. I'm happy to go ahead and approve now though if you are?

LucyGwilliamAdmin (approved):
Have tested this - platform is outputting xml files that comply with DSD and SDMX input still working well

PR #179, opened 2020-11-28
Standalone indicators

brockfanning:
Allow for indicators to have a metadata field of `standalone` which prevents them from affecting reporting status, disaggregation status, and disaggregation reports.


LucyGwilliamAdmin (approved):
Site feature branches in https://github.com/open-sdg/open-sdg/pull/988

Based off those both working and data coming through correctly, approving this

PR #180, opened 2020-11-29
Input for SDMX metadata

brockfanning:


brockfanning:
I'm wondering this PR should (optionally) grab global metadata from the UNSD API, so that it has a practical use. I'll see about adding that.

brockfanning:
@LucyGwilliamAdmin @AnnCorp Actually it's easy to grab all the metadata from the UN SDG Metadata API just by adding this class under "inputs", without any other parameters. For example:

```
inputs:
  - class: InputSdmxMeta
  - class: InputCsvData
    path_pattern: data/*.csv
  - etc...
```

It defaults to getting all the available global metadata from that API. Given that the API is only just out of the pilot stage, it doesn't yet have all indicators, but it does have some.

To make use of the imported metadata all you would need to do is adjust your data repo's _prose.yml file so that new fields are added to the "global" scope. For example:

```
- name: "SDG_GOAL"
  field:
      element: text
      label: Goal from the SDG Metadata API
      translation_key: Goal from the SDG Metadata API
      scope: global
```

This works because "SDG_GOAL" is one of the SDMX metadata concepts. The "scope: global" should cause "Goal from the SDG Metadata API" to be populated in the "Global metadata" tab in Open SDG.

For a full list of the SDMX metadata concepts, see here: https://github.com/worldbank/sdg-metadata-convert/blob/master/lib/store/concepts.json

brockfanning:
@AnnCorp @LucyGwilliamAdmin 
This one went down a bit of a rabbit hole as I've added Schema and Translation classes for the SDMX MSD. In doing so I needed to allow for multiple schemas, and also preferred to separate the reporting status types from the schema, turning it into a regular data config setting: `reporting_status_types`. I think this will be easier for countries to adjust (in contrast to that large _prose.yml file).

So even though reporting status types has nothing to do with SDMX metadata, I've included it here because the flexibility this PR offers was hampered by the old architecture.

One note: **this PR is _not_ urgent to test or merge**, because the SDMX MSD has not been released to the public yet. Much of the functionality in this PR depends on the MSD, and so I don't think we need to worry about this PR until the MSD has been released.

For fun I took a screenshot with all of the global metadata for 1.1.1 pulled from the UN metadata API. There is a bit of weirdness with font sizes, since the metadata itself contains `H1` and `H2` tags, but here it is:

global-metadata-sdmx-input-example

brockfanning:
@LucyGwilliamAdmin This is one that I think warrants discussion during a call, but I always forget. I added a "to discuss" tag to see if that helps remind me.

brockfanning:
What required discussion here was the "reporting_status_types" feature, but I'll look and see if that can be postponed to a later ticket.

brockfanning:
I've removed the "reporting_status_types" stuff from this PR and edit the comments above to reflect. Again, this PR is not urgent to review/merge yet. I'll change it to draft.

brockfanning:
We discussed going ahead with reviewing/merging even if the MSD is not final, so I'll take this out of draft state.

LucyGwilliamAdmin:
@brockfanning just to double check, what am I testing here? just pulling data from UN API for now? or do you have a SDMX metadata file I should test with?

brockfanning:
Good question - let me refresh my memory on this one and post more details.

brockfanning:
@LucyGwilliamAdmin Right now just pulling from the UN API, yes.

LucyGwilliamAdmin (approved):
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/sdmx-metadata-input/en/meta/1-1-1.json

Looks good to me:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/sdmx-metadata-input/en/meta/1-1-1.json

PR #182, opened 2020-12-14
Better handling of common SDMX data fields

brockfanning:
This allows for the possibility that SERIES can be used instead of Series, and UNIT_MEASURE can be used instead of Units, by ensuring that they will not be translated. This is meant to be used in combination with https://github.com/open-sdg/open-sdg/pull/1027

In addition this ensures that COMPOSITE_BREAKDOWN will be exempt from translations as well. This is meant to be used in combination with https://github.com/open-sdg/open-sdg/pull/1026

Here are the effects that this PR will have if merged:

## Defaults for non-disaggregation columns

Countries can specify their "non-disaggregation columns" in their data config, by doing this:

```
indicator_options:
    non_disaggregation_columns:
        - Year
        - Value
        - etc...
```

Right now, if they omit it, the defaults are: 'Year', 'Units', 'Series', 'Value', 'GeoCode', 'Observation status', 'Unit multiplier', and 'Unit measure'. With this PR, 'UNIT_MEASURE' and 'SERIES' are added to that list of defaults.

### What uses "non-disaggregation columns"?

This list of "non-disaggregation columns" is used for the following purposes:

1. This library tries to build a "headline" for each indicator, which is a set of yearly data with no disaggregation. Non-disaggregation columns are not considered when building these headlines.
2. This library builds a disaggregation report, and the non-disaggregation columns are omitted from this report.
3. This library builds an "edges" file for each indicator, which is a representation of parent/child disaggregation relationships (to show when a "child" disaggregation is only available when its "parent" disaggregation is active). Non-disaggregations are not considered when building these "edges".

So, with this PR, if the country has omitted the non_disaggregation_columns setting and is using the defaults, "UNIT_MEASURE" and "SERIES" will not affect headlines, disaggregation reports, and edges. (In the same way that "Units" and "Series" already do not affect them.)

## Translations of data column headers

Right now, the headers of data columns are always translated, except when they are "Year", "Value", or "Units". With this PR, they are always translated, except when they any of the "non-disaggregation column", or if they are "COMPOSITE_BREAKDOWN".

So, with this PR, any column headers that are non-disaggregation columns (or COMPOSITE_BREAKDOWN) will be untranslated. (In the same way that "Year", "Value", and "Units" are already untranslated.)


LucyGwilliamAdmin:
@brockfanning re. UNIT_MEASURE and SERIES, I've tested this here:
https://lucygwilliamadmin.github.io/open-sdg-site-starter-1/en/1-1-1/

I think it's a good idea for them to be able to be used instead of Units and Series, but should they behave in the same way i.e. only show as option if there's more than one option - although, this comment might be something to go in https://github.com/open-sdg/open-sdg/pull/1027 instead?

sdg/open_sdg.py
```diff
@@ -155,7 +155,10 @@ def open_sdg_indicator_options_defaults():
             'GeoCode',
             'Observation status',
             'Unit multiplier',
-            'Unit measure'
+            'Unit measure',
```
LucyGwilliamAdmin, line 158: Unit measure was something we suggested adding, as we added it to our data files in preparation for SDMX with a plan to map it to UNIT_MEASURE. I'm not sure of the best way to deal with this now as we will have 3 separate things for units: Units, Unit measure and UNIT_MEASURE.

Similarly Observation status and Unit multiplier are also common SDMX codes (in fact I think they're required) so not sure if anything needs to be done with those?

LucyGwilliamAdmin:
@brockfanning I've just realised I have dimension_map mapping UNIT_MEASURE to Units and SERIES to Series, I'm assuming this should be removed to test this?

brockfanning:
@LucyGwilliamAdmin I wanted to get this straight in my own head, so I've added more details in the PR description. Hopefully this helps explain what's going on here.

I would think that it would be useful to test without making any changes, and then also to test removing that dimension mapping.

(Though realize that the build without the dimension mapping will not display correctly in Open SDG, since Open SDG requires "Units" and "Series" specifically. Hence this other PR over there: https://github.com/open-sdg/open-sdg/pull/1027.)

sdg/open_sdg.py
brockfanning: @LucyGwilliamAdmin The question about OBS_STATUS and UNIT_MULT seems related to our SDMX output project. Our thinking there is that we should be able to produce SDMX even without specifying these required columns in the data. Instead they should be inserted automatically with defaults or configuration at the indicator level.

About the 3 separate units, yes I think this is something to discuss with your team. I see that most indicators seem to have "Unit measure", which is being ignored by Open SDG, but some have "Units", which Open SDG is displaying. Is there any reason that you don't want some indicators to display the unit toggle? (Maybe because they only have one unit?)

Also just to give you a preview of what I'm thinking: All things being equal I would prefer if Open SDG offered _no_ flexibility when it came column names. I think the platform will be more robust and easy to understand when we have clear and definite requirements about column names. And actually the platform is already like that: the column names "Year", "Value", "Series", and "Units" all have specific meanings in Open SDG and can't be spelled any other way. However as we're moving towards Open SDG as a way to input **and output** SDMX, it's become more clear to me that we should have used the SDMX codes, like "TIME_PERIOD" (or whatever), "OBS_VALUE", "SERIES", and "UNIT_MEASURE".

So my long-term goal is to switch over to those SDMX codes. In the short term though, I'm trying to make it optional so that they can be used right now (this is the motivation for this PR: https://github.com/open-sdg/open-sdg/pull/1027). Maybe in Open SDG 2.0.0 or 3.0.0 we can switch to requiring "UNIT_MEASURE" instead of "Units", for example.

LucyGwilliamAdmin:
I created a new sandpit specifically for SDMX input and I'm actually struggling to get Units and Series to show in data build at all right now: https://open-sdg-sdmx-input-sandpit.github.io/data/en/data/1-1-1.json

I'm currently pointing at this branch with no `dimension_map`

Do you know what might be causing this?

brockfanning:
@LucyGwilliamAdmin I wonder if it is the "drop_singleton_dimensions" setting, which defaults to true. Maybe try setting that to false?

LucyGwilliamAdmin:
@brockfanning that worked.

The following has been used (when dimension_map been used):
```
 dimension_map:
      UNIT_MEASURE: 'Units'
      SERIES: 'Series'
      REF_AREA|KG: ''
```

When dimension map used column names are: Year, Frequency of observation, Reporting type, Series, Reference area, Units, Value and headline reference area is null

When dimension map not used column names are: Year, Frequency of observation, Reporting type, SDG Series, Reference area, Unit of measure, Value and headline reference area is Kyrgyzstan

Should Series and SDG Series both be SERIES and Units and Unit of measure both be UNIT_MEASURE though?

Also I'm using 

```
indicator_options:
    non_disaggregation_columns:
        - Year
        - Value
        - FREQ
        - REPORTING_TYPE
        - SERIES
        - UNIT_MEASURE
 ```

brockfanning:
@LucyGwilliamAdmin I'm having trouble following the question - if you could describe the issue or goal I will try to test it out locally. It's been a while so my brain is fuzzy on this PR.

LucyGwilliamAdmin:
@brockfanning it might be that my understanding of this PR isn't right:

I'm now pointing at this branch but it seems that SERIES and UNIT_MEASURE still seem to be acting as disaggregation columns rather than non-disaggregation columns. For example:

1. There should be headline data here: https://open-sdg-sdmx-input-sandpit.github.io/data/en/headline/1-1-1.json but it's empty
2. SDG Series and Unit of measure are showing on [disaggregation report](https://open-sdg-sdmx-input-sandpit.github.io/data/disaggregations.html) and on [platform](https://open-sdg-sdmx-input-sandpit.github.io/site/1-1-1/)

However if you go to the [untranslated translations json](https://raw.githubusercontent.com/open-sdg-sdmx-input-sandpit/data/gh-pages/untranslated/translations/translations.json) and search for SDG Series it looks like SERIES is getting translated to SDG Series which I think is the cause of the problem.

brockfanning:
@LucyGwilliamAdmin I think the problem here is that the SDMX input with "import_translation_keys" enabled actually imports columns as, for example, "concept.SERIES" rather than "SERIES". This means that it doesn't get recognized as a "non-disaggregation column", and it gets translated.

This importing of translation keys is going to go away if we merge #177. I'm looking at that PR next to see if I can get it closer to being ready.

jwestw:
Hey @brockfanning 
After fixing the issue with the series appearing as disaggregation checkboxes on the Nepal platform, I now see that the series names are coming in not translated. 

image

Do you know why this is happening and is it related to this PR?

LucyGwilliamAdmin:
@brockfanning just tried this and now SERIES and UNIT_MEASURE are showing like that in the data - is this correct?

brockfanning:
@LucyGwilliamAdmin Yes I think that's what I'd expect, based on this code change.

LucyGwilliamAdmin:
Yeah, plus I checked with pointing at open-sdg branch and they're showing correctly

LucyGwilliamAdmin (approved):
SERIES and UNIT_MEASURE are now being pulled in untranslated

https://open-sdg-sdmx-input-sandpit.github.io/data/en/data/1-1-1.json

PR #183, opened 2020-12-16
Show raw value in disagg report, then all languages

brockfanning:
Fixes https://github.com/open-sdg/open-sdg/issues/1020

This makes the translation columns optional. To turn them on, the config setting is:

```
docs_translate_disaggregations: true
```

Feature branch with `docs_translate_disaggregations: true`: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/disagg-report-show-raw-value-true/disaggregations.html

Feature branch without `docs_translate_disaggregations: true`: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/disagg-report-show-raw-value/disaggregations.html

### DIFF
--- a/docs/examples/open_sdg_config.yml
+++ b/docs/examples/open_sdg_config.yml
@@ -115,6 +115,8 @@ docs_intro: This is a list of examples of endpoints and output that are
   available on this service. Click each of the links below for more information
   on the available output.
 docs_indicator_url: https://my-github-org/my-site-repository/[id]
+# Whether to provide columns for translations in the disaggregation report.
+docs_translate_disaggregations: true
 
 # Indicator downloads
 # -------------------
--- a/sdg/DisaggregationReportService.py
+++ b/sdg/DisaggregationReportService.py
@@ -147,11 +147,12 @@ class DisaggregationReportService:
         )
 
 
-    def translate(self, text, language):
+    def translate(self, text, language, group=None):
         if self.translation_helper is None:
             return text
         else:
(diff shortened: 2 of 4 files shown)


brockfanning:
@jwestw @LucyGwilliamAdmin Do you see any issues with this change? I think this is an important change for multilingual sites, where data providers may need to know which values they are allowed to put into their CSV files. With this change, data providers can use the disaggregation report as a "cheat sheet" to remind them of the allowed disaggregation values.

LucyGwilliamAdmin:
So on multilingual sites there would be an extra column which would say something like data.Sex?

brockfanning:
All sites, whether multilingual or not, would show an additional column that has whatever value is in the source data. Then the translated versions would be in additional columns to the right (even if only one, like on a monolingual site). I'll ping you on Slack with an example.

brockfanning:
@LucyGwilliamAdmin This PR now has the optional tag: `docs_translation_disaggregations`, which defaults to false.

LucyGwilliamAdmin:
@brockfanning I seem to get an error in both cases here when `docs_translate_disaggregations: true`:
https://github.com/ONSdigital/sdg-data/runs/1588815370?check_suite_focus=true#step:5:42

brockfanning:
@LucyGwilliamAdmin That looks like it had trouble downloading the GeoJSON file. You might try re-running to see if it was temporary.

LucyGwilliamAdmin (approved):
Looks good to me - links to feature branches in description

PR #186, opened 2020-12-21
Output docs accessibility fixes

brockfanning:
Fixes https://github.com/open-sdg/open-sdg/issues/1033

This PR includes these changes:

1. More descriptive labels for the Download CSV links
2. Removes 'role="grid"' in the tables. There is mention in the feedback of using similar markup to the tables in Open SDG. The tables in Open SDG are built in a pretty complex way and have been tweaked heavily for accessibility. Because these tables are built in a different way, it would take a bit of work to mimic the Open SDG tables. It can be done though if necessary. For now I've just done the specific change mentioned in the feedback: removing the role attribute.
3. More descriptive jump links at the top
4. Change order of the page titles: page title first and then branding
5. File size for CSV downloads
6. Add role=button to CSV downloads

Feature branch: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/output-docs-accessibility-fixes-take-2/disaggregations.html

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 pyyaml
 pandas
-numpy
 gitpython
 jsonschema
 requests
--- a/sdg/DisaggregationReportService.py
+++ b/sdg/DisaggregationReportService.py
@@ -264,16 +264,16 @@ class DisaggregationReportService:
     def get_disaggregation_report_template(self):
         return """
         <div class="list-group list-group-horizontal mb-4">
-            <a class="list-group-item list-group-item-action" href="#by-disaggregation">By disaggregation</a>
-            <a class="list-group-item list-group-item-action" href="#by-indicator">By indicator</a>
+            <a class="list-group-item list-group-item-action" href="#by-disaggregation">Jump to list of disaggregations</a>
+            <a class="list-group-item list-group-item-action" href="#by-indicator">Jump to list of indicators</a>
         </div>
         <div>
-            <h2 id="by-disaggregation">By disaggregation</h2>
+            <h2 id="by-disaggregation">Disaggregations</h2>
             {disaggregation_download}
             {disaggregation_table}
         </div>
         <div>
-            <h2 id="by-indicator">By indicator</h2>
+            <h2 id="by-indicator">Indicators</h2>
             {indicator_download}
             {indicator_table}
         </div>
@@ -283,16 +283,16 @@ class DisaggregationReportService:
     def get_disaggregation_detail_template(self):
         return """
         <div class="list-group list-group-horizontal mb-4">
-            <a class="list-group-item list-group-item-action" href="#values-used">Values used in disaggregation</a>
-            <a class="list-group-item list-group-item-action" href="#indicators-using">Indicators using disaggregation</a>
+            <a class="list-group-item list-group-item-action" href="#values-used">Jump to list of values used in this disaggregation</a>
+            <a class="list-group-item list-group-item-action" href="#indicators-using">Jump to list of indicators using this disaggregation</a>
         </div>
         <div>
-            <h2 id="values-used">Values used in disaggregation</h2>
+            <h2 id="values-used">Values used in this disaggregation</h2>
             {values_download}
             {values_table}
         </div>
         <div>
-            <h2 id="indicators-using">Indicators using disaggregation</h2>
+            <h2 id="indicators-using">Indicators using this disaggregation</h2>
             {indicators_download}
             {indicators_table}
         </div>
--- a/sdg/DisaggregationReportService.py
+++ b/sdg/DisaggregationReportService.py
@@ -263,17 +263,20 @@ class DisaggregationReportService:
 
     def get_disaggregation_report_template(self):
         return """
-        <div class="list-group list-group-horizontal mb-4">
-            <a class="list-group-item list-group-item-action" href="#by-disaggregation">Jump to list of disaggregations</a>
-            <a class="list-group-item list-group-item-action" href="#by-indicator">Jump to list of indicators</a>
+        <div>
+            <p>Contents:</p>
+            <ul aria-label="Table of contents">
+                <li><a aria-label="Go to list of disaggregations" href="#by-disaggregation">By disaggregation</a></li>
+                <li><a aria-label="Go to list of indicators" href="#by-indicator">By indicator</a></li>
+            </ul>
         </div>
         <div>
-            <h2 id="by-disaggregation">Disaggregations</h2>
+            <h2 id="by-disaggregation">By disaggregation</h2>
             {disaggregation_download}
             {disaggregation_table}
         </div>
         <div>
-            <h2 id="by-indicator">Indicators</h2>
+            <h2 id="by-indicator">By indicator</h2>
             {indicator_download}
             {indicator_table}
         </div>
@@ -282,17 +285,20 @@ class DisaggregationReportService:
 
     def get_disaggregation_detail_template(self):
         return """
-        <div class="list-group list-group-horizontal mb-4">
-            <a class="list-group-item list-group-item-action" href="#values-used">Jump to list of values used in this disaggregation</a>
-            <a class="list-group-item list-group-item-action" href="#indicators-using">Jump to list of indicators using this disaggregation</a>
+        <div>
+            <p>Contents:</p>
+            <ul aria-label="Table of contents">
+                <li><a aria-label="Go to list of values used in disaggregation" href="#values-used">Values used in disaggregation</a></li>
+                <li><a aria-label="Go to list of indicators using disaggregation" href="#indicators-using">Indicators using disaggregation</a></li>
+            </ul>
         </div>
         <div>
-            <h2 id="values-used">Values used in this disaggregation</h2>
+            <h2 id="values-used">Values used in disaggregation</h2>
             {values_download}
             {values_table}
         </div>
         <div>
-            <h2 id="indicators-using">Indicators using this disaggregation</h2>
+            <h2 id="indicators-using">Indicators using disaggregation</h2>
             {indicators_download}
             {indicators_table}
         </div>
--- a/sdg/DisaggregationReportService.py
+++ b/sdg/DisaggregationReportService.py
@@ -263,20 +263,20 @@ class DisaggregationReportService:
 
     def get_disaggregation_report_template(self):
         return """
-        <div>
-            <p>Contents:</p>
-            <ul aria-label="Table of contents">
-                <li><a aria-label="Go to list of disaggregations" href="#by-disaggregation">By disaggregation</a></li>
-                <li><a aria-label="Go to list of indicators" href="#by-indicator">By indicator</a></li>
+        <div role="navigation" aria-describedby="contents-heading">
+            <h2 id="contents-heading">On this page</h2>
+            <ul>
+                <li><a href="#by-disaggregation">By disaggregation</a></li>
+                <li><a href="#by-indicator">By indicator</a></li>
             </ul>
         </div>
         <div>
-            <h2 id="by-disaggregation">By disaggregation</h2>
+            <h2 id="by-disaggregation" tabindex="-1">By disaggregation</h2>
             {disaggregation_download}
             {disaggregation_table}
         </div>
         <div>
-            <h2 id="by-indicator">By indicator</h2>
+            <h2 id="by-indicator" tabindex="-1">By indicator</h2>
             {indicator_download}
             {indicator_table}
         </div>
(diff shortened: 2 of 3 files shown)


LucyGwilliamAdmin:
@brockfanning looking at this now

I'm not sure if they meant to put "Jump to" in writing or just in sr?

brockfanning:
@LucyGwilliamAdmin I'm open to whatever you prefer - they didn't request specific text, so I wrote that off the top of my head.

LucyGwilliamAdmin:
@brockfanning I've just spoke with @phillipgardiner and wondering if it might be better as more of a table of contents style like this: 


i.e.

Contents
- By disaggregation (sr text: Go to list of disaggregations)
- By indicator (sr text: Go to list of indicators)

What about something like that?

brockfanning:
@LucyGwilliamAdmin @phillipgardiner Sounds good, just pushed a commit for that change.

LucyGwilliamAdmin:
Looks good - sending back to DAC now

LucyGwilliamAdmin:
@brockfanning when I sent this back to DAC I got automatic replies - they are out of office until 4th Jan so won't receive any feedback until after then

LucyGwilliamAdmin:
Some feedback from DAC:

1. The aria-labels on table headers are still confusing because they convey an instruction. This means that when screen reader users navigate with table shortcuts so that the table headers are relayed from individual table data cells, incorrect instructions are offered.

I would suggest you implement something similar to the SDG site: https://sdgdata.gov.uk/1-1-1/

i.e. instead of 

Current Code:

```
<th data-column="1" class="tablesorter-header tablesorter-headerUnSorted" tabindex="0" scope="col" role="columnheader" aria-disabled="false" unselectable="on" aria-sort="none" aria-label="Number of indicators: No sort applied, activate to apply an ascending sort" style="user-select: none;">
   <div class="tablesorter-header-inner">Number of indicators</div>
</th>
```

Something like:

Example:

```
<div style="display: none;">
   <span id="indicators-column-sort-info">No sort applied, activate to apply an ascending sort</span>
</div>
```

And then:

```
<th data-column="1" class="tablesorter-header tablesorter-headerUnSorted" scope="col" role="columnheader" aria-sort="none" style="user-select: none;">
   <button class="tablesorter-header-inner" 
    aria-describedby="indicators-column-sort-info">Number    
    of indicators</button>
</th>
```

Instead of the table header being focusable and having an associated click event, an actionable component such as a button should be implemented inside the table header. This will ensure that interactive content as a role that assistive technology will recognise as interactive. At the moment, with voice ac software, if I say ‘Click Disaggregation’, the only thing being recognised is the ‘Download CSV of disaggregations’ button, so instead of the table sorting as expected, I accidentally trigger a download. 

2. Additionally, to ensure internal links work in IE, the target element should have a tabindex=-1.

Example:
`<h2 id="by-disaggregation" tabindex="-1">By disaggregation</h2>`

and:
`<h2 id="by-indicator" tabindex="-1">By indicator</h2>`

3. I would also say that the aria-label on the list of contents is confusing:

`<ul aria-label="Table of contents" class="">`

Semantic roles are automatically presented to users by their assistive technology. Tables have their own navigational methods with screen reading software. This is not marked-up as a <table>, so it may be confusing to tell users that it is a ‘table’, especially as screen reader users will already be relayed this as a ‘list’. Instead I would suggest something like the following:

Example:

```
<div role="navigation" aria-describedby="contents-heading">
   <h2 id="contents-heading">On this page</h2>
   <ul class="">
      <li><a href="#by-disaggregation" class="">By disaggregation</a></li>
      <li><a href="#by-indicator" class="">By indicator</a></li>
   </ul>
</div>
```

I would avoid relying on the use of aria-labels in general, and strongly recommend clear native HTML labels/text on the page along with semantic mark-up to convey structure and purpose, as this is usually sufficient for users.

phillipgardiner:
@brockfanning @LucyGwilliamAdmin small thing I have noticed, but ideally we would align numbers to the right in cells. Also just putting this link in here incase there is anything useful as there seems to be a fair bit of chat about accessability and some code etc.

brockfanning:
@LucyGwilliamAdmin Is there a feature branch for this that DAC was looking at? I just want to be sure I'm looking at the same thing. And a few comments about each item:

1. Table sorting - Would it be possible to ask DAC if they can recommend a javascript table-sorting library that does everything correctly out-of-the-box? We have a great deal of freedom here, in that we can use any front-end frameworks we want. This is in contrast to Open SDG, where we were forced to use jQuery Datatables, which required a great deal of custom code which we don't want to repeat here.
2. Adding the tabindex is easy enough.
3. A couple of comments on this navigation issue:
    * I think something got stripped out of the DAC response, by "This is not marked up as a"
    * Unless directed otherwise I will use the DAC example verbatim. I want to avoid any more back-and-forth and I don't have strong feelings about this. But let me know if you think the wording should be different from what is in the DAC example.

brockfanning:
@phillipgardiner Very useful thread! It's a lot to go through, but I see links to at least 2 libraries:
* https://frankieroberto.github.io/sortable-table/docs/
* https://ljwatson.github.io/sortable-tables/
Also there is a link to a design system component:
* https://moj-design-system.herokuapp.com/components/sortable-table

The 2 libraries are a few years old but I would be happy to try them out. I wonder if DAC could approve them first before we implement?

Also looking at the design system made me think - should we be trying to implement the Gov.uk design system? We have a lot of freedom since these "data docs" are totally separate from Open SDG. But then that makes me wonder - should we actually be trying to make the data docs mirror the look/feel of Open SDG for consistency of branding?

But anyway that will lead us down a rabbit hole for sure, and so to start, I think it would be good get DAC's blessing on a sortable-table library. Either one of their recommendation or one of the two linked above.

brockfanning:
@phillipgardiner @LucyGwilliamAdmin @jwestw Actually to save time (since this is the last thing needed for 1.2.0) I've gone ahead and implemented one of those other table-sorting libraries. I choose the https://frankieroberto.github.io/sortable-table/docs/ library. Hopefully it gets DAC's approval. I tried it with a screenreader and it seemed good to me.

Side notes:
1. I also made numeric values right-aligned, per the suggestion from @phillipgardiner 
2. I made the calls-to-action for each output more specific (instead of just "See examples")

I think that once the feature branch is updated this is ready for DAC review again. I would still be interested in knowing if DAC recommends a particular table-sorting library, but if this new one is sufficient then it's not as urgent.

LucyGwilliamAdmin:
@brockfanning @phillipgardiner I have just updated feature branch (link in description/top comment) and sent to DAC

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Hi. Some good questions above @brockfanning. Just to put the link I pasted into context. The page I linked to is where GDS discuss work in progress, before a component is officially signed off for use. As you can see the table sorting etc is not yet a formal signed off component (however these pages are always good for getting a head start / ideas). What has also happened since the creation of GDS is government departs have started creating their own asset libraries hence the 'Ministry of Justice' version (again although not strictly GDS they are required to follow similar standards etc, so useful for getting a head start and not re-creating the wheel). 

In regards to look and feel etc. I think your right we do have the freedom to move away from GDS standards. I think our approach was always if they have something lets use that as our baseline until we prove otherwise, hopefully cutting out any obvious initial mistakes (as there components go through lots of user testing etc before sign off).

Essentially if this document is sitting / linked to from Open SDG then we should look to mirror the look/feel of Open SDG, not just for consistency of branding, so as not to confuse the user (e.g. so that they know it is the same site and they have not been linked off somewhere else etc)

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Ps what would be the table experience with JavaScript turned off? Does the MOJ version require javascript too? Just thinking about the progressive enhancement work @jwestw did.

brockfanning:
@phillipgardiner Just wanted to clarify that by "we have a lot of freedom" I meant freedom of choice of software: whereas with Open SDG we are forced to use Bootstrap 3 and jQuery DataTables for table-sorting -- because we have to work with what is already there -- by contrast these "data docs" are brand new.

Regarding the progressive enhancement: anything that sorts the table columns is going to need to use javascript. The only non-javascript way to sort things dynamically is to use server-side technologies like PHP, and we don't have that option. Removing the the table-sorting would simply mean that the user cannot sort the columns on the page. If they still wanted to sort, they would need to download the CSV, open in Excel, and sort that way.

From my own experience using those tables with a country that is trying to adopt SDMX, that table-sorting is quite useful. For example under "By indicator", sometimes I want to find a particular indicator so I sort by the "Indicator" column, but other times I want to see all the indicators with exactly 1 disaggregation so I sort by the "Number of disaggregations" column ascending. Still other times I want to see the indicators with the most disaggregations, so I sort by "Number of disagregations" descending.

LucyGwilliamAdmin:
@brockfanning @phillipgardiner some feedback from DAC re. table sorting:

The table works better now, the only thing I would say is that before any sort has been applied, upon first landing on the button, i.e. in this state:   there is nothing to clearly describe the purpose of the button. 

I can see you have used a status message to inform users to the consequence of activating a button – however, they will only be relayed this after the button has been activated. Because you have the aria-sort on the table header, I would say that this status message isn’t really needed, because when the aria-sorted changes, this will be relayed to users anyway. 

Instead, I think I would implement something that made the button purpose clearer for users when they initially land on the button, such as an aria-describedby pointing at a div which said ‘Click to sort ascending/descending’. 

I haven’t got any libraries I can recommend sorry, I would honestly advise using the SDG tables as a template https://sdgdata.gov.uk/2-1-1/ I have even pointed other clients at these for a good example as I think they work very well!

brockfanning:
One thing I'm definitely hearing from this is that we need to refactor our Open SDG table code. I feel bad that people are being sent to Open SDG as an example. It's true our end result is good, but the code we had to write to get there is a mess, and spread between multiple functions and files. I don't expect anybody would be able to easily recreate what we did. (Which is why I keep saying that we don't want to do the same thing we did on Open SDG.)

So maybe the best solution here is to use the same libraries and customizations as Open SDG, but refactor the code so that it is maintainable. (Then eventually, in future releases, bring that refactoring back to Open SDG.)

I'd estimate that would take 5-10 hours. Are we OK with doing that, given that this ticket is the only one left in 1.2.0?

LucyGwilliamAdmin:
@AnnCorp what are your thoughts?

AnnCorp:
@brockfanning am fine with that depending on when you think these hours could be fitted ie by how long it would delay release of 1.2 please?

brockfanning:
@AnnCorp I could make it my top priority and get it done by the end of this week at the latest.

AnnCorp:
@brockfanning sounds good to me - will leave it to you to write out separate issue ticket for the extra work. Thank you!

brockfanning:
@LucyGwilliamAdmin This is ready for another pass by DAC for updates to the table-sorting.

Tech note: It's now using a custom library that I've pushed up here: https://github.com/open-sdg/open-sdg-table
Having this as a standalone library should make it easier to eventually refactor Open SDG to use this. And assuming this passes muster at DAC then this should be where DAC sends people for an example.

brockfanning:
@LucyGwilliamAdmin This is ready for a feature-branch update to see the more obvious sorting arrows. I think it looks much better - thanks for the suggestion @phillipgardiner.

LucyGwilliamAdmin:
I've now had feedback from DAC - if it is, like they say, a small fix, is it worth getting in?:

There is still an issue which I have just checked and it does actually happens on the main site too, apologies this wasn’t noticed previously. It is only a small fix.

On page load the code looks right:
image

However, once you activate any of the buttons to sort the column, it returns to the old version with the aria-labels on the table headers again:

image

this shouldn’t happen and does cause issues, but otherwise its great. 

I don’t think you need the aria-sort on the <span> with a role of button, as that it already on the column header:

image

But as long as the aria-label is removed from the `<th>` element this should work nicely.

brockfanning:
@LucyGwilliamAdmin I just pushed a commit that should take care of this.

LucyGwilliamAdmin:
@brockfanning have just rebuilt - I think the sorting arrows look really good!

When I click to sort aria-label doesn't appear for the column I choose to sort on but does appear for the others:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/output-docs-accessibility-fixes-take-2/disaggregations.html

image

brockfanning:
@LucyGwilliamAdmin Sorry - made another tweak and I think I've got it now. The aria-label should never display.

LucyGwilliamAdmin:
@brockfanning I've just rebuilt but still having the same issue - does the version need to be bumped again?

brockfanning:
@LucyGwilliamAdmin Arg - so sorry, I forgot to actually push the commit.

brockfanning:
Ok it's actually pushed now - sorry for the hassle.

LucyGwilliamAdmin:
Ok - this looks good. I'll send back - but I think we can just go with this now as agreed in tech call?

`aria-sort` is still appearing in the button span after a sort has been activated but I don't think an issue, just a duplication we may want to remove in future

brockfanning:
Sounds good, after merging this I'll go ahead with all the release stuff!

LucyGwilliamAdmin (approved):

phillipgardiner:
@brockfanning @LucyGwilliamAdmin Nice work 👍

PR #187, opened 2020-12-25
Input from the UN SDGs-SDMX API

brockfanning:
This is a convenience class for having an input from the UN SDGs-SDMX API. Most countries would prefer to have direct control over their data, and so this will probably not be used much. However it might be useful to quickly visualize the global data for a country.

To use it, the only parameter needed is the SDMX code for the reference area. For example, the world is "1", Afghanistan is "4", etc. For example:

```
inputs:
  - class: InputSdmxMl_UnitedNationsApi
    reference_area: 4
```

As a side note - this could also be used as a starting point to get CSV data for an Open SDG site. This would require running Python scripts locally, but could be done easily by building the data and then copying the built, untranslated CSV files into the data folder. For example, if running on the open-sdg-data-starter repo:

```
python scripts/build_data.py
cp _site/untranslated/data/*.csv data
git add data
git commit -m "CSV data from UN SDGs-SDMX API"
git push origin develop
```

Then after this point the data would be under the control of the country and could be updated/maintained as needed.


LucyGwilliamAdmin:
@brockfanning does this require DSD input?

brockfanning:
@LucyGwilliamAdmin Since the SDMX is coming from the UN database, this input automatically uses the global DSD behind the scenes.

LucyGwilliamAdmin:
@jwestw Brock suggested using this functionality to load some sample data into Nepal's platform. 

Again this would involve updating requirements.txt to point at this branch like example in #192 description.

And also the inputs section in the data config file will need to be updated to include the InputSdmxMl_UnitedNationsApi class (example in first chunk of code in description.

Brock has already found the reference area for Nepal: 524

Let me know if you have any queries!

jwestw:
Needs to point at git+https://github.com/brockfanning/sdg-build-1@un-sdmx-sdgs-api

jwestw:
Hey @brockfanning . I have built the [Nepal platform](https://open-sdg-nepal.github.io/nepal-site/) just recently. As I had noted in my last comment here, I pointed the requirements txt at this branch in order to get some of the UN SDMX data pulled across into the platform. After making (hopefully all of) the required changes, the code in the [data repo](https://github.com/open-sdg-nepal/nepal-data) won't build. The this is the [failed test.](https://github.com/open-sdg-nepal/nepal-data/pull/4/checks?check_run_id=1786572714). 

It looks like these are data validation errors. I started to look into check_data.py but @LucyGwilliamAdmin suggested that I raise this here because it seems like a problem with the source data having a trailing whitespace in some of the columns. Please can you look into this and see what we can do about it? Thanks

brockfanning:
@jwestw I just merged in the latest code into this branch to make sure it has the latest code. With this I think you should add `import_codes: true` to the input configuration. Like so:

```
inputs:
  - class: InputSdmxMl_UnitedNationsApi
    reference_area: 524
    import_codes: true
```

brockfanning:
@jwestw Also, this should be the only data input - so you should also remove the CSV input ("InputCsvData") from the list of inputs.

LucyGwilliamAdmin:
@brockfanning should translations need to be pulled from DSD separately?

brockfanning:
@jwestw If helpful, a couple more tips. Here is a config that I've found works well to get rid of the unneeded stuff:

```
inputs:
  - class: InputSdmxMl_UnitedNationsApi
    reference_area: 524
    import_codes: true
    drop_singleton_dimensions: false
    drop_dimensions:
      - UPPER_BOUND
      - LOWER_BOUND
      - UNIT_MULT
      - COMMENT_OBS
      - SOURCE_DETAIL
      - BASE_PER
      - NATURE
      - DATA_LAST_UPDATE
      - FREQ
      - REPORTING_TYPE
      - REF_AREA
      - TIME_DETAIL
      - OBS_STATUS
```

Also in order to have translated disaggregation controls, you'll want to add an SDMX translation input along with the usual sdg-translations:

```
translations:
  - class: TranslationInputSdmx
    source: https://registry.sdmx.org/ws/public/sdmxapi/rest/datastructure/IAEG-SDGs/SDG/latest/?format=sdmx-2.1&detail=full&references=children
```

jwestw:
Thank you very much @brockfanning

jwestw:
Hey @brockfanning 
After fixing the issue with the series appearing as disaggregation checkboxes on the Nepal platform, I now see that the series names are coming in not translated. 

image

Do you know why this is happening and is it related to this PR?

EDIT: Also I *think* when the series names were disaggregation checkboxes they were in human readable format.

jwestw:
Another possibly related issue is that the reference area appears in the disaggregations in some indicators. 

This is an example of what's happening in Indicator 3.2.1 currently. 

image

jwestw:
Relating to comment [above](https://github.com/open-sdg/sdg-build/pull/187#issuecomment-775101312) regarding the series names. To investigate the cause of this, Lucy suggested trying to change the DSD which the translation is reading from to [Kyrgyzstan's DSD](https://raw.githubusercontent.com/sdg-kyrgyzstan/open-sdg-data-starter/develop/SDG_DSD.KG.1.6.xml). Unfortunately this made no change to the translations. 

This has not completely ruled out the DSD being the source of the problem though. Though if the DSDs are compatible (which I am not sure of) this may be a problem in the code.

LucyGwilliamAdmin:
@jwestw @brockfanning I've just made two changed to my sandpit, which is causing my site to have the same issue with the translations that James is having

Change 1: https://github.com/open-sdg-sdmx-input-sandpit/data/commit/439ffa4d5dcd1b6c99b7e80f939f525cefcafb69
Change 2: https://github.com/open-sdg-sdmx-input-sandpit/data/commit/ef30046645d69b0da505d64c68ca67c9224a3efc

Site: https://open-sdg-sdmx-input-sandpit.github.io/site/9-1-2/

I think those changes are dependent on one another but I'll try changing one to see if can get an idea which of those 2 things is the issue

LucyGwilliamAdmin:
I've just tried changing to `import_translations_key` but no change there, although don't think it makes a diff according the the description of that parameter:
https://github.com/open-sdg/sdg-build/pull/177/files#diff-b4502d6359bf6c19f8c2c849582c8d846309038e0e2415e69053ba7ee0ee0d32R54

LucyGwilliamAdmin:
I also tried pointing directly at `brockfanning:sdmx-output-take-35` and issues still occurs then so possibly something in #177 causing this

brockfanning:
@jwestw In regards to your question about translations - that's related to some other PRs (#182 and https://github.com/open-sdg/open-sdg/pull/1027) in that they will remove the need to for the SERIES->Series and UNIT_MEASURE->Units alterations. In the meantime, you can get those translations to work by adding the dimension map to the translation as well, like this:

```
- class: TranslationInputSdmx
    source: https://raw.githubusercontent.com/sdg-kyrgyzstan/open-sdg-data-starter/develop/SDG_DSD.KG.1.6.xml
    dimension_map:
      UNIT_MEASURE: 'Units'
      SERIES: 'Series'
```

I'm not sure I'm following the other questions. My guess with the reference area issue is that it's because of the `REF_AREA|524: ''` alteration. But I haven't investigated in depth.

LucyGwilliamAdmin:
Could adding `REF_AREA|524` to the translations dimension_map fix the reference area issue?

LucyGwilliamAdmin:
@brockfanning i'm picking this up now but here's the latest

James added the dimension map to the translations input and translations are now being pulled in for Series and Units, which is great!

He also removed `REF_AREA|524: ''` from the dimension map so disagg looked like this:

He then tried having `REF_AREA|524: ''` in SDMX input dimension map and translations input map, but then it went back to looking like this:
image

I'm not too sure what the issue is as when `REF_AREA|KG: ''` is used with 1.2.0, it works as I expect i.e. when Reference area is not disaggregated, no dropdown shows in disagg section

I know we have `drop_singleton_dimensions` but trying to avoid that as it removes it completely

brockfanning:
@LucyGwilliamAdmin Is a screenshot missing above? I can give this a try locally and see what's going on. As I understand it, the goal is for the reference area dropdown to disappear completely for all indicators, correct?

brockfanning:
@LucyGwilliamAdmin Oops, nevermind - the screenshot just showed up... :)

LucyGwilliamAdmin:
@brockfanning apologies - I missed the first screenshot off so just added it.

Yeah, looking for the dropdown to completely disappear when it's only one e.g. Nepal/524 but show when there's more than one

brockfanning:
In this particular case it is always going to be 524 -- by definition -- because that is the query being sent to the UN database.

LucyGwilliamAdmin:
Ah ok, I guess that will change when they start using their own data if there's geog disaggs?

I also tested it on SDMX input sandpit and had same issue even though there were some indicators with geog disagg

brockfanning:
Actually I wonder if we could submit multiple reference areas in the query - would that be useful?

LucyGwilliamAdmin:
As it's the global DSD, I imagine it would just have country ref areas so don't think it would be useful in this case as wouldn't provide any geog disaggs

However it might be useful if for some reason someone wanted to customise their platform to allow for comparisons with other countries

brockfanning:
Just glanced at the global DSD - I don't think it has any entries for regions within Nepal, so I don't think multiple reference areas would help in this use-case.

But I wanted to stress that this input is not something a country would want to continue to use if they also want control of their data. So, if they wanted to add disaggregation by region, they would probably want to:

1. Temporarily use this input
2. Grab the untranslated CSV output that this input creates
3. Commit it to their repo
4. Start using those CSV files as input, _instead_ of this UN database input
5. At that point they'd be free to add any disaggregations they like to those CSV files.

LucyGwilliamAdmin:
Yes, so this Nepal case was just to add some sample data and test this functionality - they have actually done SDMX data modelling with UNSD so they'll be using that when we hand over the platform to them.

But that sounds like a good method for those who want to use data from API but want to customise it

brockfanning:
Ok sounds good. Do you think we need to pursue that multiple-ref-area idea, or keep this simple for now?

LucyGwilliamAdmin:
I think we could keep simple for now with just the one but keep in mind as possibility in future if we end up going down that route

LucyGwilliamAdmin:
I could perhaps open a new issue as don't think related to this but do you think there's a way to fix the Reference area dropdown issue?

brockfanning:
If I understand right, the dropdown issue is that if there is a CSV column without any content, it still shows up as an empty dropdown. Do I have that right? There are a few different ways to consider fixing that:
1. In sdg-build - make sure that all empty columns are dropped before output is generated
2. In open-sdg - Add code to avoid making dropdowns if there are no options
3. In docs - Remind users not to include columns that have no content

LucyGwilliamAdmin:
Yeah, that's right - are any of those options able to be done locally e.g. using JS hook?

Also, I've just checked another indicator on the Nepal platform:

1.1.1 doesn't seem to have the same issue, even though Reference area is present in data with all nulls - [site](https://open-sdg-nepal.github.io/nepal-site/1-1-1/), [data](https://open-sdg-nepal.github.io/nepal-data/en/data/1-1-1.json)

brockfanning:
@LucyGwilliamAdmin Before we go too deep into troubleshooting that, I'd suggest you try switching the Nepal site to the 1.3.0-dev branch, since we just merged that series/unit revamp, in case that fixes it.

LucyGwilliamAdmin:
Ok, so `drop_singleton_dimensions` defaults to true, which explains the difference between the Nepal platform and Kyrgyzstan's platform - I was wondering why Kyrgyzstan weren't having this issue.

I think I could just make the same change for now - I was trying to avoid this as changes data on input which I was trying to avoid

LucyGwilliamAdmin:
@brockfanning sorry submitted comment at same time. Do we need to merge this one as well to test that out?

I have no issue with that, as input from API seems to be working great

LucyGwilliamAdmin (approved):
Approving this as it works great - Nepal platform currently pulling data from API

brockfanning:
@LucyGwilliamAdmin Sounds good - actually I was talking about the 1.3.0-dev branch of open-sdg - but happy to have this merged too, thanks!

LucyGwilliamAdmin:
@brockfanning upgrading to 1.3.0-dev has fixed it - thanks so much for your help!

jwestw:
Hey @brockfanning I have a few questions as I am writing up notes about this process which may later get integrated into the docs. Would appreciate your help. 

- is there a website that has some general info about the API being used here?
- who prepared the data that's on the API?
- Is the data available just related to SDGs?
- where can reference areas be looked up?

I guess the answer the first one may provide answers to the others. Thank you

brockfanning:
@jwestw 

* My technical resource was [this manual](https://unstats.un.org/sdgs/files/SDMX_SDG_API_MANUAL.pdf) which I found on the [SDMX Working Group page](https://unstats.un.org/sdgs/iaeg-sdgs/sdmx-working-group/).
* This data was submitted to the UN by "custodian agencies" who are in charge of gathering the global data for their indicators.
* Yes this database is just for the SDGs.
* The authoritative listing of reference areas is contained in the "global DSD", which you can get from the SDMX Working Group page, with either "SDG DSD Matrix v1.3" (an Excel file) or "Global SDG DSD v1.3" (an XML file).

jwestw:
Brilliant. Thank you @brockfanning

PR #188, opened 2020-12-27
Allow missing values in reporting status extra fields

brockfanning:
Fixes https://github.com/open-sdg/open-sdg/issues/1010

This just needs a translation. I have the translation key as `status.not_specified`, thinking the English translation will be "Not specified", but it can be changed as needed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 pyyaml
 pandas
-numpy
 gitpython
 jsonschema
 requests
--- a/sdg/stats.py
+++ b/sdg/stats.py
@@ -41,6 +41,7 @@ def reporting_status(schema, all_meta, extra_fields=None):
     ]
     # Convert that into a dataframe.
     meta_df = pd.DataFrame(rows, index=all_meta.keys())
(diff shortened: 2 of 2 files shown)


LucyGwilliamAdmin (approved):
This works well:
https://lucygwilliamadmin.github.io/open-sdg-site-starter-1/en/reporting-status/

image

Side note: how would I translate indicator_type here again? Would it come under status translations?

brockfanning:
@LucyGwilliamAdmin I believe the best way is to specify the "translation_key" properties in the schema file. For example here is how Kazakhstan is translating one of their custom metadata fields: https://github.com/kazstat/sdg-data-kazstat/blob/develop/_prose.yml#L70-L88

Do you think "Not specified" is the best English here? If so I will add it to SDG Translations so it can be translated into the other languages.

LucyGwilliamAdmin:
I think "Not specified" works - clear and true. The other one I thought of was "Other" but that may lead people to think that the country specified it as "Other!

LucyGwilliamAdmin:
Also just tried the prose thing for the translation - that worked. Thanks :)

PR #189, opened 2020-12-29
Allow the export filename to be configured

brockfanning:
Fixes #679

The data config setting to test here is: `indicator_export_filename`, and should be the desired filename without the extension. For example:

`indicator_export_filename: uk_sdg_indicators`

### DIFF
--- a/docs/examples/open_sdg_config.yml
+++ b/docs/examples/open_sdg_config.yml
@@ -132,3 +132,10 @@ indicator_downloads:
     output_folder: data-csv
     # The "indicator_id_pattern" is optional, default below:
     indicator_id_pattern: indicator_(.*)
+
+# Indicator export (zip file)
+# ---------------------------
+# This controls the filename of the zipped file of indicator data.
+# The file extension (.zip) is not needed. The defualt, if omitted,
+# is below:
+indicator_export_filename: all_indicators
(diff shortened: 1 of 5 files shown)


LucyGwilliamAdmin:
@brockfanning have just tried testing this but it didn't seem to work. Here's my data config change:
https://github.com/ONSdigital/sdg-data/blob/configurable-indicator-download-filename/config_data.yml#L94

and see zip file info here:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/configurable-indicator-download-filename/open-sdg-output.html

brockfanning:
@LucyGwilliamAdmin I hadn't thought this one all the way through - but I believe I've caught everything now. This will also need a [corresponding PR in open-sdg](https://github.com/open-sdg/open-sdg/pull/1061) or else the link on site will be broken.

LucyGwilliamAdmin:
@brockfanning thanks - i've just rebuilt the feature branch:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/configurable-indicator-download-filename/open-sdg-output.html

I'm wondering if the name of the json file should be change as well?

LucyGwilliamAdmin (approved):
Ah ok cool - happy to approve then

PR #190, opened 2021-01-01
Data schema and data packages

### LINKED ISSUE #19: Support csvw
CSV For the web https://www.w3.org/TR/tabular-data-primer/

Both for input and output. This allows us to store data attributes (column information) alongside the data.

brockfanning:
Fixes #19 

This brings in the frictionless and csvw libraries and uses them as a basis for data schema input/validation, as well as a data packages output and a CSVW output.

At the same time, data schema is used to allow SDMX output to be "constrained" by the DSD. In other words, to only output data that is compliant with the DSD.

(These two items may seem completely different, and are. They're just two features made possible by having a "data schema".)

## Usage

In order to use this, the following needs to be in the requirements.txt file:

```
git+https://github.com/brockfanning/sdg-build-1@data-schema-2
```

### Constraining data in SDMX output

To try out the data constraining feature, set "constrain_data" to true, in the SDMX config. For example:

```
sdmx_output:
  constrain_data: true
```

### Data package output

Without any configuration, this will add an output for [data packages](https://specs.frictionlessdata.io/data-package/). The schema for these data packages is "inferred" from the source data itself.

The output can be customized with the `datapackage` config setting, which can contain any parameters to pass to an OutputDataPackage object. The available parameters are "package_properties", "resource_properties", "field_properties", and "sorting". Details are in the class definition.

#### Controlling data package output per indicator

The parameters mentioned above can also be placed in a "datapackage" metadata field, to have the same effect for particular indicators.

### CSVW output

Optionally, if there is a `csvw` config setting, CSVW output will be produced. The setting can simply be `csvw: true` to use the defaults, or it can be parameters to be passed to an OutputCsvw object. The available parameters include all those for data packages, plus "common_properties", "table_schema_properties", "column_properties", and "at_properties". Details are in the class definition.

#### Controlling CSVW output per indicator

The parameters mentioned above can also be placed in a "csvw" metadata field, to have the same effect for particular indicators.

### Difference between data packages and CSVW

As detailed above the OutputCsvw takes all the same parameters as the OutputDataPackage, plus some extras. This is because OutputCsvw actually inherits from OutputDataPackage. Behind the scenes of OutputCsvw a data package is being created, and then converted into CSVW. The data package parameters ("package_properties", "resource_properties", and "field_properties") are applied to the generation of the data package, and the CSVW parameters ("common_properties", "table_schema_properties", and "column_properties" are applied to the CSVW after conversion.

### Sorting

Both outputs take a "sorting" parameter which can be either "alphabetical" or "default". Counterintuitively the default is not "default", and instead it's "alphabetical". This is to keep the existing sorting behavior in Open SDG, which is alphabetical, but eventually we probably want to change the default to "default", as would be expected.

#### Why sorting matters

Eventually, the sorting of the fields/values in these outputs can/will be used in Open SDG to control the sorting of the disaggregation dropdowns and checkboxes.

## Controlling the schema per indicator

If full control over the schema is needed, instead of "inferring" schema from the data, it can be provided as YAML files in a subfolder. For example a "data-schema" folder:

* data-schema
    * 1-1-1.yml
    * 1-2-1.yml
    * etc..

These schema files should conform to the [Table Schema](https://specs.frictionlessdata.io/table-schema/) spec. For example:

```
fields:
  - name: Year
    type: integer
  - name: Sex
    type: string
    constraints:
      enum:
        - Male
        - Female
        - ""
  - name: Age
    type: string
    constraints:
      enum:
        - Under 16
        - 16 to 24
        - 25 to 49
        - 50 to 64
        - 65 to 74
        - 75 and over
        - ""
  - name: Value
    type: number
```

This path of the files to use is controlled with a config setting of "data_schema". For example:

```
data_schema:
  source: data-schema/*.yml
```

### Data package output, schema inferred from source data

To generate only data packages, with schema inferred from the source data, no configuration is needed. Simply use this branch.

### Data package output and CSVW output, schema inferred from source data

```
csvw: true
```

### Data package output and CSVW output, schema inferred from source data, non-alphabetical sorting

```
datapackage:
  sorting: default
csvw:
  sorting: default
```

### Data package output and CSVW output, schema inferred from source data, custom properties in CSVW

```
csvw:
  common_properties:
    "dc:description": "My description"
  table_schema_properties:
    aboutUrl: https://example.com
  column_properties:
    Year:
      propertyUrl: https://example.com
```


brockfanning:
@jwestw I would appreciate your feedback on this. I've been looking at the frictionless-py library, and this could be a pretty easy avenue towards CSVW. You can test this as a data branch, and then see the data packages in the output docs:
screenshot-127 0 0 1_8080-2021 01 01-16_42_07

brockfanning:
@jwestw In particular, I'm wondering if this [Frictionless data package spec](https://specs.frictionlessdata.io/data-package/) is at all what your team is looking for. I suspect that it is not, as the more I read I see that this Frictionless spec is different from the [W3C CSVW spec](https://www.w3.org/TR/tabular-data-primer/). In that case, I might want to rename this to `OutputFrictionlessDataPackage` instead, and make a different one for the W3C spec, like maybe `OutputCsvwDataPackage`.

However this same approach could be used for a CSVW output - so long as there is a library that can "infer" a CSVW schema in the same way that the Frictionless library does. Do you have any recommendations? A quick search finds these:

* https://github.com/cldf/csvw
* https://github.com/bloomberg/pycsvw

Both of those say that they deviate from the W3C CSVW spec. But maybe that's unavoidable. If we need to support the W3C spec instead of the Frictionless spec, my hope is that one of these libraries has the ability to look at tabular data and "infer" the schema.

jwestw:
Hi @brockfanning . It looks like you've come across most of the resources that I found at the beginning of my CSVW project. 
 I had looked at the Frictionless library, the Bloomberg library, the [cldf/csvw ](https://github.com/cldf/csvw)library, and the CLARIAH/COW stuff too. 

The main reason I didn't use any of those was because our [Connected Open Government Statistics](https://gss.civilservice.gov.uk/guidance/the-gss-data-project/) (COGS) team have their own set of tools for achieving CSVW and RDF output, and my work has been in conjunction with them.

Apart from that I can't remember *exactly* why I didn't go for each but I think it was something like:
* Frictionless: Didn't seem to adhere to W3C CSVW standards. Though it does something similar. Also, I seemed to think it was proprietary at the time. 
* Bloomberg: Looked really promising but all CLI based. Looking at it again now though it looks really good. 

Since then I have been working with the COGS team. The most recent version has is on a fork here: https://github.com/mikeAdamss/sdg-csv-data-filler/tree/POC3_mike

Check the out folder in that branch which contains the meta data json files that have been generated.

brockfanning:
@jwestw Thanks - I think you had showed me this before, but it makes a lot more sense now that I'm into the details with this.

It looks like the Bloomberg repo is unmaintained - or at least unchanged since it was posted - whereas the cldf/csvw library is more active. I posted a question there and got a very quick response - I think they will be able to add a Frictionless-to-CSVW converter, which should make it very easy to use this code to also produce CSVW. In the meantime, I'll rename the class in this PR.

I also looked at the COGS CSVWMapper code. If I'm reading it right, that does allow the generation of the JSON metadata from a CSV file. I'll keep that in mind if the Frictionless-to-CSVW approach doesn't work out.

By the way - a side question about the sdg-csv-data-filler code: Have you considered doing the CSV alterations (filling gaps, changing headers, changing cells, etc.) on the input instead of during the CSVW output? With sdg-build you can alter the data during the input -- [see here for an example](https://github.com/sdg-kyrgyzstan/open-sdg-data-starter/blob/develop/scripts/check_and_build_data.py#L6-L12). The reason I bring that up is that presumably those alterations generally improve the data quality - in which case it would be nice to display those improved data on the NRP. This way you'd have improved data on the NRP, and you would not need to do anything special during the CSVW output.

jwestw:
Using Frictionless and cldf/csvw sounds like it might be a good strategy. I guess if you can work with those and you feel they are well supported then go with that. 
The overlap in our work does make me wonder though if it is worth speaking to our COGs team to see if their tools would work for your tasks and if they will be continue to be supported going forward. 

Also regarding applying data transformations on the data input side, I think this makes sense if it is allowed., as it would make the CSV and CSVW data more uniform. I did not think this was an option though for some reasons. To be discussed with Lucy/Ann.

brockfanning:
@jwestw I think this is ready to test, though I suspect there will be a lot of back-and-forth while we refine the output. When you can, it would be helpful if you could try it out, and let me know of the shortcomings in the CSVW output. Then we can tweak the the code as needed. Usage details are in the description.

brockfanning:
@jwestw It might also help if I post the current output. I'm trying to replicate the output [here](https://github.com/mikeAdamss/sdg-csv-data-filler/blob/POC3_mike/out/indicator_11-7-1.csv-metadata.json). I'm using the data config to get some of the way there:

```
csvw:
  table_schema_properties:
    aboutUrl: http://gss-data.org.uk/data/specify-path-here-please/{year}/{country}/{observation_status}/{unit_multiplier}/{unit_measure}/{geocode}
  column_properties:
    Year:
      propertyUrl: http://purl.org/linked-data/sdmx/2009/dimension#refPeriod
      valueUrl: http://reference.data.gov.uk/id/{+year}
    Value:
      propertyUrl: http://gss-data.org.uk/def/measure/count
  at_properties:
    id: http://gss-data.org.uk/data/specify-path-here-please#tables
  common_properties:
    "prov:hadDerivation":
        "@id": http://gss-data.org.uk/data/specify-path-here-please#dataset
```

And here is the CSVW output:

```
{
    "@context": [
        "http://www.w3.org/ns/csvw",
        {
            "@language": "en"
        }
    ],
    "@id": "http://gss-data.org.uk/data/specify-path-here-please#tables",
    "dc:conformsTo": "data-package",
    "dc:identifier": "11-7-1",
    "dc:replaces": "{\"profile\": \"data-package\", \"resources\": [{\"path\": \"data.csv\", \"stats\": {\"hash\": \"a5afa4005696b1dd6a283c4f4748328a\", \"bytes\": 290, \"fields\": 7, \"rows\": 4}, \"control\": {\"newline\": \"\"}, \"encoding\": \"utf-8\", \"dialect\": {}, \"schema\": {\"fields\": [{\"name\": \"Country\", \"type\": \"string\", \"constraints\": {\"enum\": [\"\", \"England\", \"Scotland\", \"Wales\"]}, \"title\": \"Country\"}, {\"name\": \"GeoCode\", \"type\": \"string\", \"title\": \"GeoCode\"}, {\"name\": \"Observation status\", \"type\": \"string\", \"title\": \"Observation status\"}, {\"name\": \"Unit measure\", \"type\": \"string\", \"title\": \"Unit measure\"}, {\"name\": \"Unit multiplier\", \"type\": \"string\", \"title\": \"Unit multiplier\"}, {\"name\": \"Value\", \"type\": \"number\", \"title\": \"Value\"}, {\"name\": \"Year\", \"type\": \"integer\", \"title\": \"Year\"}]}, \"name\": \"data\", \"profile\": \"tabular-data-resource\", \"scheme\": \"file\", \"format\": \"csv\", \"hashing\": \"md5\", \"compression\": \"no\", \"compressionPath\": \"\", \"query\": {}}], \"name\": \"11-7-1\", \"title\": \"Average share of the built-up area of cities that is open space for public use for all, by sex, age and persons with disabilities\"}",
    "dc:title": "Average share of the built-up area of cities that is open space for public use for all, by sex, age and persons with disabilities",
    "prov:hadDerivation": {
        "@id": "http://gss-data.org.uk/data/specify-path-here-please#dataset"
    },
    "tables": [
        {
            "tableSchema": {
                "aboutUrl": "http://gss-data.org.uk/data/specify-path-here-please/{year}/{country}/{observation_status}/{unit_multiplier}/{unit_measure}/{geocode}",
                "columns": [
                    {
                        "datatype": "string",
                        "name": "Country",
                        "titles": "Country"
                    },
                    {
                        "datatype": "string",
                        "name": "GeoCode",
                        "titles": "GeoCode"
                    },
                    {
                        "datatype": "string",
                        "name": "Observation status",
                        "titles": "Observation status"
                    },
                    {
                        "datatype": "string",
                        "name": "Unit measure",
                        "titles": "Unit measure"
                    },
                    {
                        "datatype": "string",
                        "name": "Unit multiplier",
                        "titles": "Unit multiplier"
                    },
                    {
                        "datatype": "number",
                        "propertyUrl": "http://gss-data.org.uk/def/measure/count",
                        "name": "Value",
                        "titles": "Value"
                    },
                    {
                        "datatype": "integer",
                        "propertyUrl": "http://purl.org/linked-data/sdmx/2009/dimension#refPeriod",
                        "valueUrl": "http://reference.data.gov.uk/id/{+year}",
                        "name": "Year",
                        "titles": "Year"
                    }
                ]
            },
            "url": "data.csv"
        }
    ]
}
```

LucyGwilliamAdmin:
@brockfanning I just want to double check, are we expecting for this one to go into 1.3.0?

brockfanning:
@LucyGwilliamAdmin I was thinking we might discuss this in a tech call. But here's my thinking: Ideally we could work on this iteratively, something like this:

1. Feature branch is created to demonstrate the output
2. Someone gives feedback on the output (requests, issues, etc.)
3. Code is tweaked according to the feedback
4. Feature branch is updated, then back to step 2.

Given that we haven't really started that yet, I'd expect this may not be ready for 1.3.0, but I suppose there is the chance that it is perfect as is. ;)

jwestw:
I have made a [data feature branch repo](https://github.com/ONSdigital/sdg-data/tree/data-schema-2). I have included `csvw: true` and also pointed the requirements/.txt at git+https://github.com/brockfanning/sdg-build-1@data-schema-2

The branch has built with no problems. Where should I be seeing CSVW data output?

brockfanning:
@jwestw I would take a look in the "data docs" for the feature branch (the data build).

jwestw:
You can see some sample output here: http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/data-schema-2/csvw.html

brockfanning:
Cool - so my main questions are:
1. Is there anything missing from the JSON file that should be added?
2. Is there anything wrong about the JSON file that should be fixed?
3. Is there anything unnecessary in the JSON file that should be removed?

jwestw:
Okay, I will try to answer those questions. 

I am also considering validating with this: https://csvlint.floop.org.uk/ (Thanks @LucyGwilliamAdmin Lucy)

jwestw:
I am currently comparing these metadata json file for 16-9-1:
- [this created from my project with COGS ](https://raw.githubusercontent.com/ONSdigital/sdg-csv-data-filler/POC3/out/proportion-of-children-under-5-years-of-age-whose-births-have-been-registered-with-a-civil-authority-by-age.csv-metadata.json)
- [this created by Brock's script ](http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/data/data-schema-2/en/csvw/16-9-1/datapackage.json)

jwestw:
### Validation using CSVlint:

When validating the json for V2 with 16-9-1.csv:

image

The warnings about invalid name, may be related to my comment above about the names not being lowercased.

When attempting to validate V1 json however, **CSVlint crashes**. I would post the output here, but it's long.

jwestw:
I am going to re-configure my branch with the similar properties to those [suggested above by Brock](https://github.com/open-sdg/sdg-build/pull/190#issuecomment-757055224) to see if I can get the output to be closer to V1. 

These are the settings so far: 


```yaml
csvw:
  table_schema_properties:
    aboutUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler/proportion-of-children-under-5-years-of-age-whose-births-have-been-registered-with-a-civil-authority-by-age/{year}/{sex}/{unit_measure}/{observation_status}/{unit_multiplier}/{value}
  column_properties:
    Year:
      propertyUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler/proportion-of-children-under-5-years-of-age-whose-births-have-been-registered-with-a-civil-authority-by-age#dimension/year
      valueUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler#concept/year/{year}
    Sex:
      propertyUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler/proportion-of-children-under-5-years-of-age-whose-births-have-been-registered-with-a-civil-authority-by-age#dimension/sex
      valueUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler#concept/sex/{sex}          
    Unit measure:
      propertyUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler/proportion-of-children-under-5-years-of-age-whose-births-have-been-registered-with-a-civil-authority-by-age#dimension/unit-measure
      valueUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler#concept/unit-measure/{unit_measure}
    Observation status:
      propertyUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler/proportion-of-children-under-5-years-of-age-whose-births-have-been-registered-with-a-civil-authority-by-age#dimension/observation-status
      valueUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler#concept/observation-status/{observation_status}
    Unit multiplier:
      propertyUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler/proportion-of-children-under-5-years-of-age-whose-births-have-been-registered-with-a-civil-authority-by-age#dimension/unit-multiplier
      valueUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler#concept/unit-multiplier/{unit_multiplier}
    Value:
      propertyUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler/proportion-of-children-under-5-years-of-age-whose-births-have-been-registered-with-a-civil-authority-by-age#dimension/value
      valueUrl: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler#concept/value/{value}"
  at_properties:
    id: http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler/proportion-of-children-under-5-years-of-age-whose-births-have-been-registered-with-a-civil-authority-by-age#tables
  common_properties:
    "prov:hadDerivation":
        "@id": http://gss-data.org.uk/data/gss_data/sdg/sdg-csv-data-filler/proportion-of-children-under-5-years-of-age-whose-births-have-been-registered-with-a-civil-authority-by-age#dataset
```

brockfanning:
@jwestw If I understand right, each indicator would need to have unique URLs for its fields? So for example, the "propertyUrl" of "Year" for indicator 1.1.1 would be different from the "propertyUrl" of "Year" for indicator 1.2.1?

jwestw:
Hey Brock, thanks for your input and questions. 

> So for example, the "propertyUrl" of "Year" for indicator 1.1.1 would be different from the "propertyUrl" of "Year" for indicator 1.2.1?

The script created with our COGS/IDP team, the json for each indicator has unique urls for each column, even if it has the same name, e.g. "Year" in each dataset. I am only deducing this from the json that results from running the code though I am not sure why this is; I can ask if needed. And whether the propertyUrl in the CSVW format needs to be unique on a per dataset per column basis or not is unknown to me, but I suspect not. 

I notice that w3.org has propertyUrls that can be linked to, like: https://www.w3.org/2006/time#year

brockfanning:
@jwestw Yes maybe it would be helpful to ask the COGS/IDP team why each indicator/column combination needs a unique URL. Maintaining content at all of those URLs seems like a monumental job (like 1000+ assuming roughly 4 columns per indicator) - but maybe there is something I'm missing.

The challenge here is that we're developing a feature but we don't know how it will be used - so it's difficult to decide on what configuration options it should have. To take two possibilities:

1. Users need to be able to set these URLs for any indicator/column combination: In this case we need the configuration options to be at the indicator level.
2. Users only need to be able to set these URLs for any column - to be shared across all indicators. In this case we only need the configuration options at the site level.

The difference between those options is huge - since maintaining configuration per indicator is significantly more effort (for the user) than site-wide.

And even if the requirement is number 1 above, there may be a way to help by using placeholders. For example, maybe the site-wide URL for the "Year" column could be: https://example.com/[indicator]/Year - where "[indicator]" gets automatically replaced by the indicator ID. For that matter, maybe it would be possible to have a site-wide URL like https://example.com/[indicator]/[column], where [indicator] gets replaced by the indicator ID and [column] gets replaced by the column name.

jwestw:
@brockfanning I have asked those questions, so let's see what they come back with.

robons:
> @jwestw Yes maybe it would be helpful to ask the COGS/IDP team why each indicator/column combination needs a unique URL. Maintaining content at all of those URLs seems like a monumental job (like 1000+ assuming roughly 4 columns per indicator) - but maybe there is something I'm missing.

So the CSV-Ws we're generating on the IDP-D platform don't all have one distinct URI per column. Many of the datasets we're importing utilised these shared URI definitions, but there can be a bit of a cost associated with this. If we know that the columns truly represent the same concept there there's no problem, however we do come across a number of data publishers who might name things similarly, but the concepts actually represent slightly different things; in this case it can often be safest to define separate URIs for each column in each dataset, but have them show a hierarchical relationship to some parent URI which groups these concepts together. 

To make onboarding of users simpler, we've decided that it's easiest to use our tools to get users on the platform with distinct URIs for each column/dataset combination and to later start encouraging them to figure out what the true globally-defined URI should be later. We've found that defining what these common URIs should be is a task which will likely never be complete where we have repeated iterations towards better and better defined global definitions.

It really depends what kind of a platform you're building. If you're building for a single project where all concepts are well-defined then defining these global URIs should be quite simple - and this is definitely the best approach you can take so it's clear how you compare data between datasets. However, if you're building a tool which is to be used across multiple projects which might have similar overlapping concepts, then defining unique URIs for each individual column/dataset combo might be the best place to start.
> 
> The challenge here is that we're developing a feature but we don't know how it will be used - so it's difficult to decide on what configuration options it should have. To take two possibilities:
> 
>     1. Users need to be able to set these URLs for any indicator/column combination: In this case we need the configuration options to be at the indicator level.
> 
>     2. Users only need to be able to set these URLs for any column - to be shared across all indicators. In this case we only need the configuration options at the site level.
> 
> 
> The difference between those options is huge - since maintaining configuration per indicator is significantly more effort (for the user) than site-wide.
> 
> And even if the requirement is number 1 above, there may be a way to help by using placeholders. For example, maybe the site-wide URL for the "Year" column could be: [https://example.com/[indicator]/Year](https://example.com/%5Bindicator%5D/Year) - where "[indicator]" gets automatically replaced by the indicator ID. For that matter, maybe it would be possible to have a site-wide URL like [https://example.com/[indicator]/[column]](https://example.com/%5Bindicator%5D/%5Bcolumn%5D), where [indicator] gets replaced by the indicator ID and [column] gets replaced by the column name.

There are some pre-existing ontologies which have URIs which can and should be reused to define certain concepts - and will make your life easier in using them. Examples of ones which might be useful for you are:

* [reference.data.gov.uk URIs for date times and intervals] (https://github.com/epimorphics/IntervalServer/blob/master/interval-uris.md)
* [Units of measure](http://www.ontology-of-units-of-measure.org/) 
* [WikiData](https://www.wikidata.org/wiki/Wikidata:Main_Page) is often a good standard of URIs which can be used in this fashion to describe certain entities.

But there are many other sources of URIs which can be used to defined concepts and entities out there on the web. There are a *huge* number of them.

Basically, in the IDP-D project we're trying to decide which pre-defined ontologies to use and how to make the whole process of guiding statistical data publishers towards using these standardised definitions easier for them. It's a big task.



I hope what I've written helps and doesn't just open up a new rabbit hole for you.

brockfanning:
@robons @jwestw Thank you for the insights, this is very helpful. And in my opinion that gives us a clear direction - we are building towards a single project with well-defined concepts: the UN's Sustainable Development Goals. So I think that means it is OK to keep our configuration options at the site-level. Eg, the users specify one site-wide propertyUrl for "UNIT_MEASURE" and it applies to all "UNIT_MEASURE" columns across all CSV files.

brockfanning:
As discussed on the last tech call, we'll try to support per-indicator configurations, but will not require them, and will have site-wide configurations as a fallback.

brockfanning:
@LucyGwilliamAdmin @jwestw @AnnCorp I wanted to explain my recent commits here - I know we had talked about postponing this work. I needed this for a different contract, so did this work on my own time. I'm now at a stopping place here, so if you want to review that is fine - no problem if not.

One new feature that I've added here (and documented above in the description) is the ability for the SDMX output to be "constrained" by the DSD. For example, to filter out data that is non-compliant out of the SDMX output. (I know that SDMX is unrelated to CSVW, but actually it's just another feature that's made possible by having a "data schema", so I threw it in here.) This will be useful to countries that would like their SDMX output to be compatible with the global DSD.

Also, the last feature mentioned - having the ability to have per-indicator configuration, is also now possible (and documented in the description above).

LucyGwilliamAdmin:
@jwestw I have just taken a look at priorities and following Brock's explanation of what is needed, can you take a look at this after you have finished what you are currently working on please?

Can you look at whether or not its throwing any errors or making any changes to current platform?

I will look at SDMX side of things

jwestw:
Assigned this to Lucy after we talked about it.

LucyGwilliamAdmin:
@brockfanning is there anything specific I need to do for SDMX side of things? Other than put in some data that doesn't comply with DSD and make sure it's not in output

brockfanning:
@LucyGwilliamAdmin That sounds good, along with this data-config:

```
sdmx_output:
  constrain_data: true
```

LucyGwilliamAdmin:
@brockfanning ok thanks - so some non-compliance scenarios I'm thinking of testing are:
- series code that isn't in global DSD
- disaggregation that isn't in global DSD
- breakdown that isn't in global DSD
- disaggregation that is in DSD but not asked for in specific series
- breakdown that is in DSD but not asked for in specific series

Is there anything else you can think of?

brockfanning:
@LucyGwilliamAdmin Those sound good. I don't think all of those are relevant to this PR, but it's fine to test them. I believe the only type of issue that this PR fixes is if a disaggregation column has some value that is not part of its corresponding code list. From your list I think that is the first two:

* series code that isn't in global DSD
* disaggregation that isn't in global DSD

I think all the others (if I'm understanding right) are already being automatically fixed.

brockfanning:
@LucyGwilliamAdmin Sounds like a lot for me to look into, thanks for the feedback. In the meantime could you link me to the additional series code and disaggregation in the CSV files?

LucyGwilliamAdmin:
Retest:
L_POV_1 is still showing but the only series that are showing are those in which an Age breakdown exists in the DSD

 test|works
-|-
series code that isn't in global DSD|No
disaggregation that isn't in global DSD|Yes - series doesn't show at all (although the file does get listed [here](https://onsdigital.github.io/sdg-sdmx-data/sdmx-output.html) even though it doesn't exist)
breakdown that isn't in global DSD|Yes

brockfanning:
Thanks @LucyGwilliamAdmin I think I've got a fix for that issue. I've added some code that will ignore the metadata-field-as-a-default approach, if that default value is invalid.

LucyGwilliamAdmin:
@brockfanning re. the non-DSD series being ignored, the data is still showing but instead SERIES is _T:
https://onsdigital.github.io/sdg-sdmx-data/sdmx/1-1-1.xml

The other issue has been resolved though and filtered out files no longer show in the examples:
https://onsdigital.github.io/sdg-sdmx-data/sdmx-output.html

brockfanning:
@LucyGwilliamAdmin Gotcha. I don't quite have my brain wrapped around this one, so I'll talk it out. This comes up when a country has specified a default value by having a metadata field on the indicator (eg, a metadata field called "SERIES"). The point of this data constraint stuff is really to filter out stuff that is in the _data_ -- in practical terms, to filter any problems that might have been unintentionally caused by the data providers. I'm not sure if data providers would be often setting the default values for an indicator. If a country wants their data constrained to the DSD, but then puts an invalid value _as a default_, that would be pretty unusual, right?

Now that I've talked that out, I'm wondering if maybe we should trust the user to choose valid defaults, and not worry about constraining them. Put another way, maybe the user-defined default values should always come through, regardless of whether they're valid? If so, then I could probably revert my last change. If not, then I'll need to do some more refactoring.

FYI I didn't change anything regarding the output docs examples, at least intentionally...

LucyGwilliamAdmin:
@brockfanning yes - that completely makes sense that countries shouldn't be setting those incorrectly and we can leave it up to them to control that.

brockfanning:
@LucyGwilliamAdmin Ok, given that, I've reverted my last commit. Would it be possible for you to do a test where there is a series column in the data, and some of the rows have an invalid series code? (rather than the default-in-metadata approach)

LucyGwilliamAdmin:
@brockfanning all good!
When all are non-DSD series, nothing shows and when some are non-DSD series, those values don't show

LucyGwilliamAdmin:
@brockfanning Also, just to confirm, the data packages jsons are loading for me now

Is there anything else you think I should test before approving?

brockfanning:
@LucyGwilliamAdmin Sounds great, I think this is all set.

LucyGwilliamAdmin (approved):
Approving following above comments

PR #193, opened 2021-01-10
Fix path for metadata output docs

brockfanning:
Fixing a pretty straightforward typo.


brockfanning:
I'll go ahead and merge this, I don't think this needs testing because it's a pretty obvious bug.

PR #194, opened 2021-01-21
Allow missing metadata to be altered

### LINKED ISSUE #54: Allow metadata alteration even if input has no metadata
In some cases, the input only contains data, but we would still like to have some control over the metadata. Is there some way to change [this function](https://github.com/open-sdg/sdg-build/blob/master/sdg/inputs/InputBase.py#L185) so that it will apply the metadata alterations to an empty dict, if necessary?

brockfanning:
Fixes #54

### DIFF
--- a/sdg/inputs/InputBase.py
+++ b/sdg/inputs/InputBase.py
@@ -208,9 +208,11 @@ class InputBase:
         ---------
         meta : dict or None
         """
-        # If empty or None, do nothing.
         if not meta or meta is None:
-            return meta
+            if len(self.meta_alterations) > 0:
+                meta = {}
+            else:
+                return meta
         for alteration in self.meta_alterations:
             meta = alteration(meta)
         return meta


LucyGwilliamAdmin:
I pointed at this branch, then removed the metadata file for 1.1.1. I then used alter_meta to change the graph title for all indicators, and it worked for 1.1.1:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_LucyGwilliamAdmin-patch-2/1-1-1

However, the link didn't show on the goals page:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_LucyGwilliamAdmin-patch-2/1/

Would this be in another PR though?

brockfanning:
@LucyGwilliamAdmin My guess is that it is because the indicator_name field is missing. So I think that is OK. You could confirm that be setting indicator_name in the metadata alter as well.

LucyGwilliamAdmin (approved):
Looks good

http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_LucyGwilliamAdmin-patch-2/1/

PR #196, opened 2021-01-21
Use only 2 segments for the auto target id

### LINKED ISSUE #195: Automatic target id is incorrect
The auto-generated target id is showing up with three "segments", like "1-1-1". It needs to only be two segments, like "1-1".

brockfanning:
Fixes #195

### DIFF
--- a/sdg/Indicator.py
+++ b/sdg/Indicator.py
@@ -172,7 +172,7 @@ class Indicator:
         string
             The target id, dot-delimited.
         """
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin (approved):
Looks good:
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_fix-automatic-target-id/1/

PR #198, opened 2021-01-23
Allow numeric disaggregaton

brockfanning:
Fixes https://github.com/open-sdg/open-sdg/issues/327

This does two things:
1. Prevent Python exceptions when a disaggregation column contains numeric values
2. Ensure the data output treats disaggregation columns as strings, even if they are numeric

## Testing

To help with testing I created a pull-request on the Open SDG project to see what happens when a disaggregation is numeric: https://github.com/open-sdg/open-sdg/pull/1072

You can see that the first automated test succeeded, because it was pointing at this `brockfanning/sdg-build-1@allow-numeric-disaggregation` branch. Then I reverted and pointed back at the `open-sdg/sdg-build@1.3.0-dev` branch, and as expected the tests failed.

Manual tests are still a good idea, because this is a potentially impactful change. But hopefully this "failure demonstration" provides some extra assurance that this fixes the problem.

### DIFF
--- a/sdg/DisaggregationReportService.py
+++ b/sdg/DisaggregationReportService.py
@@ -121,7 +121,7 @@ class DisaggregationReportService:
         string
             The title converted into a unique *.html filename
         """
-        slug = prefix + slugify(title)
+        slug = prefix + slugify(str(title))
         if slug in self.slugs:
             slug = slug + '_'
         if len(slug) > 100:
--- a/sdg/OutputDocumentationService.py
+++ b/sdg/OutputDocumentationService.py
@@ -298,8 +298,8 @@ class OutputDocumentationService:
 
     def write_disaggregation_value_detail_page(self, info):
         service = self.disaggregation_report_service
-        disaggregation = info['disaggregation']
-        disaggregation_value = info['name']
+        disaggregation = str(info['disaggregation'])
+        disaggregation_value = str(info['name'])
         filename = info['filename']
 
         df = service.get_disaggregation_value_dataframe(info)
--- a/sdg/inputs/InputBase.py
+++ b/sdg/inputs/InputBase.py
@@ -61,6 +61,8 @@ class InputBase:
         cols = df.columns.tolist()
         cols.pop(cols.index('Year'))
         cols.pop(cols.index('Value'))
+        for col in cols:
+            df[col] = df[col].map(str)
         cols = ['Year'] + cols + ['Value']
         return df[cols]
 
@@ -77,7 +79,7 @@ class InputBase:
         Dataframe
             The same dataframe with rearranged columns
         """
-        return df.replace([None, ""], np.NaN)
+        return df.replace([None, "", "nan"], np.NaN)
 
 
     def fetch_file(self, location):


jwestw:
Hey @brockfanning I have a question. When casting Series values to str in InputBase.py, could this overlook problems in a Series that has mixed data types that should not have?

jwestw:
@brockfanning @LucyGwilliamAdmin 

I wonder if you have any pointers on how I would get started testing https://github.com/open-sdg/sdg-build/pull/198

I see the PR you've made with numeric values in https://github.com/open-sdg/open-sdg/pull/1072 , but beyond this how do I get started testing this?

- Do I need to test this locally? 
- I think I will need to build some data with numeric values. Would it be best to clone the UK data repo, change some of the data, and then build it while pointing at your brockfanning:allow-numeric-disaggregation branch? If so, that means pointing the requirements text in ./scripts/ at git+git://github.com/brockfanning/sdg-build-1@allow-numeric-disaggregation (I note it's sdg-build-1, right?)

LucyGwilliamAdmin:
@jwestw in response to your questions:

1. You are correct that you need to make changes to data for this but no need to clone or build locally - this can be done on UK feature branch. You still do what you mentioned though i.e. point at brockfanning/sdg-build-1@allow_numeric_disaggregation.
2. Start by building the data feature branch - this will test this sdg-build PR
3. Once you can see that this is working you can test the open-sdg PR by a) pointing at your data feature branch and b) pointing at corresponding open-sdg branch and also carrying out any other changes mentioned in the open-sdg PR (if there are any)

brockfanning:
@jwestw Regarding the casting question: I guess the assumption of this PR is that everything besides "Year" and "Value" can be safely converted to a string. If you can think of any issues that could cause, definitely let us know. Nothing came to my mind but I could be missing something.

@jwestw @LucyGwilliamAdmin Since this particular issue is not already affecting the UK (since there is no numeric disaggregation) do you think it would be helpful to do an additional feature branch which introduces numeric disaggregation, but which does *not* point at this PR? (In order to demonstrate the problem?)

LucyGwilliamAdmin:
@brockfanning I agree also having a feature branch which doesn't point at this PR branch is a good idea

jwestw:
@brockfanning I am not sure if these would matter at all, just thoughts:
- A col that is supposed to hold numeric disaggregations contains the value "1   " (with trailing whitespaces). Converting to a string, will maintain the whitespace. 
- If numeric disaggregations  ever go into the the thousands, commas could be used, so "1,200" for example. Casting that to a string would maintain the comma, whereas casting to an int would cause an error I think. Does this matter?

jwestw:
I made 2 branches with the same data, where 1-1-1 and 3-1-1 both contained numeric disaggregations 

To the [numeric_disaggs_nofix](https://github.com/ONSdigital/sdg-data/tree/numeric_dissag_nofix) I did nothing else, but to [numeric_disaggs_fix](https://github.com/ONSdigital/sdg-data/tree/numeric_dissags_fix) I updated requirements.txt to point at the branch of this PR. 

On branch built the other didn't:

image

Here are the last few lines of the Traceback of the [failed test](https://github.com/ONSdigital/sdg-data/runs/1919970607?check_suite_focus=true), for reference. 


image

brockfanning:
@jwestw Thanks, those results support my PR, which I always appreciate. ;)

Regarding the casting: The idea here is that disaggregations should always be treated as strings, no matter what. For example, if the value in the CSV is `1`, it will be treated as `"1"`. And as you mention `1 ` will be treated as `"1 "`. Regarding the thousands, `1000` would be treated as `"1000"`, and `1,200` would be treated as `"1,200"`. I think this should all be fine, since there is no reason to maintain any "integer-ness" of disaggregation values, that I'm aware of. This PR is intended to allow builds to complete successfully even when disaggregation values are numeric.

jwestw:
@brockfanning That all makes sense. Thanks for making that clear!

Also, I thought over the weekend, that the point at which you're casting to string, the object is already in a DataFrame, so the datatype of each Series will have already been decided by Pandas. So if any of the concerns I had raised were relevant that would be at an earlier stage, data import, stage.  

Part of the reason I was raising this is because I thought it was better to numeric data as numeric datatypes for consistency, memory and speed. Though memory and speed might not be relevant here as our data is so small. And speed might be doubly irrevant because we are not performing operations on the data.

jwestw (approved):
Approving based on tests and conversation

PR #199, opened 2021-01-25
Allow non-numeric goal ids

brockfanning:


LucyGwilliamAdmin:
@brockfanning is this one ready for review?

brockfanning:
@LucyGwilliamAdmin I meant to add details here - this is one of three related PRs across open-sdg, sdg-build, and jekyll-open-sdg-plugins - all for the same purpose. I'm leaving these all in draft mode for now, as this is part of work I'm doing for Vanuatu which is using a non-SDG framework which needs goal ids like "ECO1" and "ECO2". So nope - no review needed.

PR #201, opened 2021-01-30
SDMX output - variable type fix

brockfanning:
I noticed when using the SDMX output that errors happened in some cases when the value was an integer. Changing to a string here is an easy fix.


LucyGwilliamAdmin:
@brockfanning do you have an example of a value causing an error? just so I can test

brockfanning:
@LucyGwilliamAdmin I haven't been able to identify a cause for the errors. It looks like the fix for dimensions has been merged in a separate PR, so now this only has the fix for attributes. I think it's probably a good idea - all XML properties need to be strings anyway, by definition, so we may as well explicitly do this rather than relying on third-party code.

PR #203, opened 2021-02-07
Support booleans and add safety code in metadata mappings

brockfanning:
Support "true" and "false" in CSV/Excel metadata input. This is mainly to support the "source_active_1" metadata field in Open SDG.

Also this adds safety code for the case where the metadata mapping's key is equal to its value - which was causing that field to be deleted from the build. It's an unlikely edge case but is avoidable with this simple fix.

### DIFF
--- a/sdg/inputs/InputMetaFiles.py
+++ b/sdg/inputs/InputMetaFiles.py
@@ -36,6 +36,7 @@ class InputMetaFiles(InputFiles):
         for inid in indicator_map:
             meta = self.read_meta(indicator_map[inid])
             self.apply_metadata_mapping(meta)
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin (approved):
When pointing at 1.3.0-dev (no sources show as boolean not allowed):
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_1.3.0-meta-test/1-1-1/

When pointing at this branch (source shows as boolean allowed):
http://sdgdev-813006012.eu-west-1.elb.amazonaws.com/datapreview_meta-file-input-tweaks/1-1-1/

PR #210, opened 2021-02-28
Allow schema_file parameter to work

brockfanning:
This fixes a regression bug in 1.3.0-dev, which is preventing the "schema_file" config option from working. This bug probably was not obvious because in most cases with Open SDG the schema_file is set to the default, which is "_prose.yml". But the schema_file config option is supposed to be able to specify a different file name. For example, we're supposed to be able to rename it to "foo.yml" and then set `schema_file: foo.yml` in the data config. This PR is to restore that functionality, which got broken by a recent PR.

### DIFF
--- a/sdg/open_sdg.py
+++ b/sdg/open_sdg.py
@@ -89,8 +89,6 @@ def open_sdg_build(src_dir='', site_dir='_site', schema_file='_prose.yml',
         inputs = open_sdg_input_defaults()
     if translations is None:
         translations = open_sdg_translation_defaults()
-    if schema is None:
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin (approved):
When pointing at 1.3.0-dev and renaming _prose.yml to foo.yml and updating config file, I get this error:
https://github.com/ONSdigital/sdg-data/runs/2005654987?check_suite_focus=true

When pointing at this branch and renaming _prose.yml to foo.yml and updating config file, metadata is built using new schema file.

PR #212, opened 2021-02-28
Allow data for a single indicator to come from multiple inputs

brockfanning:
This is a low-priority feature without a clear user requirement, but I wanted to post this in case it's interesting. I think the main use-case for this would be something like:

* The main data input is difficult to edit and lacks some disaggregation (such as data from a third-party web service or something like that)
* The developer would like to have a local CSV file with additional data rows to add the necessary disaggregation
* So the developer adds a CSV input, expecting the CSV data to be "concatenated" onto the other data

Since it's a minor code change I thought it would be worth posting. This would definitely need testing though, in case it has any unforeseen consequences.

### DIFF
--- a/sdg/Indicator.py
+++ b/sdg/Indicator.py
@@ -105,6 +105,7 @@ class Indicator(Loggable):
 
         if self.has_data():
             self.data = pd.concat([self.data, val], sort=False)
+            self.enforce_column_order()
         else:
             self.data = val
         self.set_headline()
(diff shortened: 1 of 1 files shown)


brockfanning:
The user requesting this doesn't need it anymore, so I'll close this to save development time for more important features.

brockfanning:
Re-opening this since it was discussed that it may be useful. I suspect it needs to be tweaked to work with the latest code though.

PR #217, opened 2021-03-07
Always set all dimensions in SDMX output

brockfanning:
In SDMX all of the dimensions are required for each series. So the output here should always set each dimension, regardless of whether there is a value or not. If not, it should always use the default value.

### DIFF
--- a/sdg/outputs/OutputSdmxMl.py
+++ b/sdg/outputs/OutputSdmxMl.py
@@ -168,9 +168,11 @@ class OutputSdmxMl(OutputBase):
             # Skip the TIME_PERIOD dimension because it is used as the "observation dimension".
             if dimension.id == 'TIME_PERIOD':
                 continue
-            value = row[dimension.id] if dimension.id in row else self.get_dimension_default(dimension.id, indicator)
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
@brockfanning can latest be pulled in from 1.4.0 so I can make use of the updated sdmx library please?

brockfanning:
@LucyGwilliamAdmin I think it should be there - it's not showing in Github but I've merged the latest code to this branch.

LucyGwilliamAdmin:
@brockfanning looks good to me:

https://onsdigital.github.io/sdg-sdmx-data/sdmx/3-2-1.xml

LucyGwilliamAdmin (approved):

brockfanning:
@LucyGwilliamAdmin Since we're about to push a hotfix release 1.3.1, we have the choice of whether to put this in 1.3.1 or wait for 1.4.0. Thoughts?

LucyGwilliamAdmin:
@brockfanning I think this is fine to go into 1.3.1 hotfix since it's kind of a bug

PR #218, opened 2021-03-10
SDMX output: StructureSpecific

### LINKED ISSUE #205: SDMX output - support StructureSpecific
I think right now our SDMX output is always GenericData. But since the UNSD SDG Lab requires StructureSpecific, we should allow for that option. In think it will be an easy thing to add thanks to the sdmx Python library we're leveraging.

brockfanning:
Fixes #205 

With version 2.3.0 of the sdmx library, we can now produce StructureSpecific output. This PR makes this the default for Open SDG output, and also makes it an optional parameter.

### DIFF
--- a/sdg/open_sdg.py
+++ b/sdg/open_sdg.py
@@ -335,6 +335,7 @@ def open_sdg_prep(options):
             output_folder=options['site_dir'],
             translations=options['translations'],
             indicator_options=options['indicator_options'],
+            structure_specific=True,
             **options['sdmx_output']
         ))
 
--- a/sdg/open_sdg.py
+++ b/sdg/open_sdg.py
@@ -329,13 +329,14 @@ def open_sdg_prep(options):
 
     # Add SDMX output if configured.
     if 'sdmx_output' in options and 'dsd' in options['sdmx_output']:
+        if 'structure_specific' not in options['sdmx_output']:
+            options['sdmx_output'] = True
         outputs.append(sdg.outputs.OutputSdmxMl(
             inputs=inputs,
             schema=schema,
             output_folder=options['site_dir'],
             translations=options['translations'],
             indicator_options=options['indicator_options'],
-            structure_specific=True,
             **options['sdmx_output']
         ))
 
--- a/sdg/outputs/OutputSdmxMl.py
+++ b/sdg/outputs/OutputSdmxMl.py
@@ -10,7 +10,7 @@ from sdmx.model import (
     Key,
     AttributeValue,
     Observation,
-    GenericTimeSeriesDataSet,
+    StructureSpecificTimeSeriesDataSet,
     DataflowDefinition,
     Agency
 )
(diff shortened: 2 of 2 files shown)


brockfanning:
@LucyGwilliamAdmin Not urgent, but I thought you might be interested in this. You can see this in action here: https://kazstat.github.io/sdg-data-kazstat/sdmx-output.html

LucyGwilliamAdmin:
@brockfanning this is great! You can get a file with all data in as well??

brockfanning:
@LucyGwilliamAdmin You mean the "all.xml"? Yep that is already a feature of the OutputSdmxMl class.

LucyGwilliamAdmin:
@brockfanning I think you might have said somewhere else, but is structure specific required for SDG Lab?

brockfanning:
@LucyGwilliamAdmin Yes it is.

LucyGwilliamAdmin:
@brockfanning what's the best way to test this? I've merged these changes into my sdmx-mapping-2 branch and now using that on UK sandpit and structure specific xmls are being outputted.

I've just ran it through web SDMX converter to validate and all fine, so now about to try loading to SDG Lab to see if successful

LucyGwilliamAdmin:
Ok so 5/5 individual files and the all file passed the validation. 5/5 individual files could be uploaded to SDG Lab but when I try to upload the all file to SDG Lab, it doesn't work but I don't know if it's supposed to?

Aside from that, is there anything else I should test?

LucyGwilliamAdmin:
@brockfanning Daniel has just confirmed that SDG Lab accepts both single and multi-series files.

I think I see the problem and possibly not related to this - I've just revisited my multi-series file and it seems some parts of header are missing e.g. ID, Prepared, Sender:

https://onsdigital.github.io/sdg-sdmx-data/sdmx/all.xml

brockfanning:
@LucyGwilliamAdmin Ah OK, makes sense. I think we can fix that in a separate PR.

LucyGwilliamAdmin:
@brockfanning how do I configure whether or not structure specific is used?

I assumed I would need to add `structure_specific` option but I'm getting this error and I've tried setting it to False and false (both gave the same error):
https://github.com/ONSdigital/sdg-sdmx-data/runs/2142432493?check_suite_focus=true#step:5:15

brockfanning:
@LucyGwilliamAdmin For Open SDG I assumed that we would always want StructureSpecific, so I hardcoded that. I can remove the hardcoding if you think it should be optional.

LucyGwilliamAdmin:
@brockfanning I think it's fine to have that as only option - I just seen in description that it had been added as an optional parameter so was just trying to test it

LucyGwilliamAdmin (approved):
Approving as platform is outputting structure specific SDMX files which are passing validation and are being accepted by SDG Lab

brockfanning:
@LucyGwilliamAdmin Should I remove the concept_map stuff before merging?

Also I think I can tweak it so that structure_specific can be set to false without causing an error - with true still being the default.

LucyGwilliamAdmin:
@brockfanning yes please remove the mapping stuff - forgot about that

I'm fine with it not being optional but if it's a quick thing, then I don't see the harm in it making it so - what are your thoughts?

brockfanning:
@LucyGwilliamAdmin Making optional is easy - I've just pushed that, and also removed the concept_map code for now.

PR #220, opened 2021-03-12
Resolve issue when there are multiple indicator_downloads

brockfanning:


LucyGwilliamAdmin (approved):
This has fixed the issue - do you think ok to go into 1.3.0?

brockfanning:
Sure, I'd be OK with that. This is not the kind of thing that might have side-effects. I'll change the base for the PR before merging.

PR #223, opened 2021-03-16
Sdmx mapping test

LucyGwilliamAdmin:


LucyGwilliamAdmin:
@brockfanning just opened this PR so I could show you my changes - I might be completely off but is this going in the right direction?

brockfanning:
@LucyGwilliamAdmin Yes that seems like a good way to go. The parameter "sdmx_mapping" is a little redundant, since it's in an SDMX-specific class. Maybe just "mapping" or "concept_map"?

Does it correctly ignore cell values if they are not mapped?

LucyGwilliamAdmin:
@brockfanning yes, happy to change mapping name.

Do you mean if there's a cell value in csv data file that doesn't exist in mapping, then it just gets left as it is?

brockfanning:
> 
> 
> @brockfanning yes, happy to change mapping name.
> 
> Do you mean if there's a cell value in csv data file that doesn't exist in mapping, then it just gets left as it is?

Yep, that's what I was getting at - so that users don't have to map every single column and value.

LucyGwilliamAdmin:
@brockfanning I needed to make a couple of changes to this - it wasn't quite right. How much work do you think would be required to also get it into sdg-build rather than it sitting on a branch that I point at?

brockfanning:
@LucyGwilliamAdmin The main thing I think it lacks right now is documentation. This mapping would need to be in a very specific format, right? CSV file, with certain column headers? That would need to be documented somewhere.

I think you are best situated to decide if this mechanism (a CSV file) is the ideal solution. If you think it might be more cumbersome than some other approach, we might want to investigate more. But if you think this is the best, that's cool too.

brockfanning:
@LucyGwilliamAdmin Between YAML and CSV it seems like CSV might be best, since this type of file would be edited by data providers, probably, and should be relatively non-technical. I guess just imagine how you envision the UK maintaining it going forward, and what they would prefer?

LucyGwilliamAdmin:
@brockfanning I've just spoke to our data team and they want to move forward with CSV but with an ID column added - I don't think that would make a difference though.

One thing I'm wondering is about the column headers - I guess we want those to make as much sense as possible.

- Do you think we should rename them to what I've put?
- Do we want to make the columns changeable e.g. have some config settings which specifies what column the SDMX concepts, which specifies the CSV column name, etc.?

brockfanning:
Todo: investigate "transcoding" in SDMX Converter.

brockfanning:
@LucyGwilliamAdmin In SDMX Converter the transcoding is done in a sheet in the Excel file with the following structure:

Text|Dimension|Value
---|---|---
my non-sdmx value|MY_DIMENSION|my_sdmx_code

Do you think we could standardize around that structure for the CSV file here?

brockfanning:
@LucyGwilliamAdmin For the purposes of renaming columns, what if the "Dimension" column were left blank? For example:

Text|Dimension|Value
---|---|---
my non-sdmx value|MY_DIMENSION|my_sdmx_code
my non-sdmx column||my_sdmx_column

Or might that cause confusion?

LucyGwilliamAdmin:
@brockfanning I think we could definitely standardise around that structure for the mapping. I'm not too sure about the column mapping side of things.

I can't really see it being an issue but might need to change code here as think currently mapping depends on all 4 columns being available

brockfanning:
@LucyGwilliamAdmin Ok, well in the end the main thing is if we settle on a format that is intuitive and documented. Following some sort of standard is nice, but in this case I don't think it should be the main concern. I do think that the Text,Dimension,Value columns are pretty simple and intuitive.

Just throwing out an idea - Do you think it might be better to have two files - one for code mappings and one for column mappings? (eg, "Text,Dimension,Value" for code mappings and "Text,Value" for column mappings?)

FYI I'm also imagining that what we use here could -- eventually -- be used in other inputs/outputs. In theory, any input/output could benefit from the ability to convert columns/codes based on a file. But that can be tackled much later. I'm just saying that I like this idea and think it could eventually be used across the library.

LucyGwilliamAdmin:
I actually quite like the idea of having two separate mappings for columns and codes, and could possibly be simpler to maintain.

Agree that later it would be good to allow mappings across library

LucyGwilliamAdmin:
Closing as just opened new PR #236

PR #224, opened 2021-03-16
Compatibility with latest version of sdmx library

brockfanning:
If this works, it should be released as a hotfix (1.3.1) so this will be merged into master.


brockfanning:
@LucyGwilliamAdmin If you want to try this out, I've merged it into your sdmx-mapping branch, and pushed it to a new branch: sdmx-mapping-2.

LucyGwilliamAdmin (approved):
@brockfanning this worked great thank you

PR #237, opened 2021-03-29
Bugfix - structure-specific default

brockfanning:


brockfanning:
I'm 99% sure this is fine so I'll go ahead and merge, since it's already bugged.

PR #245, opened 2021-04-13
Do not sort datapackage schemas that were intentionally specified

brockfanning:
This tweak prevents the datapackage sorting logic from applying to datapackage schemas that were intentionally provided. The assumption is that if the schemas were intentionally provided then their sorting should be left alone.


brockfanning:
This is a bugfix that is tricky to test, so I'll go ahead and merge. We'll be testing this functionality very soon in https://github.com/open-sdg/open-sdg/pull/1045 so we should notice any issues then.

PR #250, opened 2021-04-18
Use TIME_PERIOD instead of TIME_DETAIL in SDMX output

brockfanning:
When I wrote the SDMX output code I was not clear on the difference between TIME_PERIOD and TIME_DETAIL - now I realize that TIME_PERIOD was more appropriate.

### DIFF
--- a/sdg/outputs/OutputSdmxMl.py
+++ b/sdg/outputs/OutputSdmxMl.py
@@ -137,7 +137,7 @@ class OutputSdmxMl(OutputBase):
                 'Value': 'OBS_VALUE',
                 'Units': 'UNIT_MEASURE',
                 'Series': 'SERIES',
-                'Year': 'TIME_DETAIL',
+                'Year': 'TIME_PERIOD',
             })
 
             if self.constrain_data:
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
@brockfanning I'm just trying to understand the difference between `TIME_PERIOD` and `TIME_DETAIL` here: https://unstats.un.org/sdgs/files/SDG-DSD-Guidelines.pdf

It looks like `TIME_PERIOD` is high level, whilst `TIME_DETAIL` is lower level - this is the example provided:
for period ‘2001-2003’ TIME_PERIOD would be 2002 but the actual dates --2001-2003-- would be expressed here

So definitely think `TIME_PERIOD` should be used here, but how does TIME_DETAIL come into it - would a country just add another column to the input called `TIME_DETAIL`?

LucyGwilliamAdmin (approved):

brockfanning:
@LucyGwilliamAdmin My understanding is that TIME_DETAIL is optional, and intended for cases where the data may be about a range of years. TIME_PERIOD is required to be a single year, but TIME_DETAIL can be any free text. So if a data set is about 2015-2020, then the TIME_PERIOD could be set to 2015 (or 2020) and the TIME_DETAIL could be "2015-2020".

PR #257, opened 2021-05-21
Vary by language in some method caches

brockfanning:
This fixes a bug where the GeoJSON output doesn't build correctly in the non-default languages. This bug has the effect in Open SDG of preventing maps from working in non-default languages, so this would be an important fix to get in for 1.4.0, I think.

### DIFF
--- a/sdg/Indicator.py
+++ b/sdg/Indicator.py
@@ -35,7 +35,7 @@ class Indicator(Loggable):
         self.set_headline()
         self.set_edges()
         self.translations = {}
-        self.serieses = None
+        self.serieses = {}
         self.data_matching_schema = {}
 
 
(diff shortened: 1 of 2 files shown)


LucyGwilliamAdmin:
@brockfanning was this bug reported by someone? Just asking before I try to recreate and test fix

brockfanning:
@LucyGwilliamAdmin No, I only noticed this while trying to set up maps for a multilingual site (Cambodia).

LucyGwilliamAdmin:
@brockfanning ok, thanks - is this working for them? or should I do separate test?

brockfanning:
@LucyGwilliamAdmin Yes, it's working well, here's an example with a map: https://sdg-cambodia.github.io/site/2-2-1/

LucyGwilliamAdmin (approved):
Looks good on Cambodia example

PR #258, opened 2021-05-25
Replace values individually to avoid exceptions

brockfanning:
This is another pass after #252 seems to have caused some issues, perhaps just in Python 3.8.


brockfanning:
I'll go ahead and merge - as the current branch is throwing errors, at least in Python 3.8.

PR #259, opened 2021-05-26
Metadata input from "SDG Metadata" repositories

brockfanning:
This allows the input of metadata from a source structured according to the [SDG Metadata project](https://github.com/worldbank/sdg-metadata). This structure can be described as:

* One main folder containing multiple subfolders
* Each subfolder is a language code
* Inside language folders are YAML files named according to the indicator id (eg 1-1-1.yml)

In other words, this is nearly identical to the InputYamlMeta class which we already have, except for these differences:

1. In this new class, the default language is in a subfolder just like the other folders. By contrast, in InputYamlMeta, the default language is in the main folder.
2. This new class allows reading directly from a Git repository, in additional to a local folder.

### DIFF
--- a/sdg/inputs/InputSdgMetadata.py
+++ b/sdg/inputs/InputSdgMetadata.py
@@ -1,0 +1,144 @@
+# -*- coding: utf-8 -*-
+
+import os
+import stat
+import shutil
+import yaml
+from git import Repo
+from sdg.inputs import InputBase
+
+class InputSdgMetadata(InputBase):
+    """This class imports metadata from SDG Metadata (or similar) repos.
+
+    The "SDG Metadata" style can be described like this:
+    1. A Git repository (passed as the "source" parameter)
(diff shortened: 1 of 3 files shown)


brockfanning:
@LucyGwilliamAdmin This is ready for review. This can be seen in use [here](https://github.com/sdg-cambodia/data/blob/develop/config_data.yml#L51-L53). With that usage, this class is reading in metadata from folders which are managed by Weblate, allowing Weblate to act as a sort-of "metadata editor" for the non-default languages.

LucyGwilliamAdmin:
@brockfanning does this require any further testing from me? or is what you've done with Cambodia suffice?

brockfanning:
@LucyGwilliamAdmin No I don't think you need to test it, since it's working in that case.

LucyGwilliamAdmin (approved):

PR #260, opened 2021-06-04
Add TIME_PERIOD to data schema from DSD

brockfanning:
We recently switched from using TIME_DETAIL to TIME_PERIOD, and now the constrain_data option doesn't work in the SDMX output, because TIME_PERIOD is intentionally skipped in the schema, which causes it to think that all rows are invalid. I can't remember the reason that TIME_PERIOD was skipped before. In a quick test this seems to work for me.


brockfanning:
Closing - duplicate of #249

PR #264, opened 2021-06-17
Do hardcoded column changes before column map

### LINKED ISSUE #263: No data outputted - SDMX output
When I use `constrain_data: true`, no data at all is being output:
https://lucygwilliamadmin.github.io/sdg-sdmx-data/sdmx/all.xml

Looking at the [data for 1.1.1](https://lucygwilliamadmin.github.io/sdg-sdmx-data/en/data/1-1-1.json), the [column mapping](https://github.com/ONSdigital/sdg-sdmx-data/blob/develop/column_mapping.csv), the [code mapping](https://github.com/ONSdigital/sdg-sdmx-data/blob/develop/code_mapping.csv) and the data config, at least 1.1.1 should be showing right?

brockfanning:
Fixes #263 

This changes the point at which the hardcoded column mappings are performed.

### DIFF
--- a/sdg/outputs/OutputSdmxMl.py
+++ b/sdg/outputs/OutputSdmxMl.py
@@ -159,9 +159,6 @@ class OutputSdmxMl(OutputBase):
             indicator = self.get_indicator_by_id(indicator_id)
             data = indicator.data.copy()
 
-            self.apply_column_map(data)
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin (approved):
Great!

Thanks for fixing so quickly @brockfanning

PR #265, opened 2021-06-17
SDMX output by goal

brockfanning:
Because SDMX data and metadata can be combined arbitrarily, it's not too hard to group them by goal. This may be useful for some cases. One immediate use-case is that the "all.xml" files (when full) are too big to be uploaded into UNSD's SDG Lab -- the upload fails. This means each indicator needs to be uploaded separately. Grouping them by goal makes this process much easier (17 uploads instead of 200+).

[Here's an example of this in action](http://brock.tips/fcdo/burundi/sdg-build/sdmx-output.html).

### DIFF
--- a/sdg/outputs/OutputSdmxMl.py
+++ b/sdg/outputs/OutputSdmxMl.py
@@ -146,7 +146,9 @@ class OutputSdmxMl(OutputBase):
             return status
 
         all_serieses = {}
+        all_serieses_by_goal = {}
         all_metadata_serieses = []
+        all_metadata_serieses_by_goal = {}
         metadata_template = Template(self.get_metadata_template())
         dfd = DataflowDefinition(id="OPEN_SDG_DFD", structure=self.dsd)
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
@brockfanning should have asked before I started testing but is this ready for review?

My findings so far:

- Goal files have no validation errors
- However, it does seem that some of the series that are in the goal 1 file don't belong to goal 1 indicators (haven't checked other goals). E.g.
  - SH_H2O_SAFE (6.6.1)
  - IT_MOB_OWN (5.b.1)

brockfanning:
@LucyGwilliamAdmin I thought it was ready, but that does seem wrong. I'll look into that, thanks for catching it.

brockfanning:
@LucyGwilliamAdmin I'm realizing that the discrepancy you noticed is a bug in the source data of that example, but not a bug from this PR. Here is another example using Kyrgyzstan's data which should be more correct: http://brock.tips/fcdo-kg-data/

LucyGwilliamAdmin (approved):
Looks good based on Kyrgyzstan goal output

PR #270, opened 2021-07-21
Handle special case of id/name columns in GeoJson output

brockfanning:
Fixes https://github.com/open-sdg/open-sdg/issues/1271

The specific id column name of 'geocode' and the name column of 'name' cause problems. This avoids those problems.


LucyGwilliamAdmin (approved):
Tested this here: https://lucygwilliamadmin.github.io/open-sdg-simple-starter-1/8-3-1/

Works well!

PR #271, opened 2021-07-21
Allow a dtype parameter in CSV data input

brockfanning:
Fixes https://github.com/open-sdg/open-sdg/issues/1271

This allows a dtype parameter to be passed into the InputCsvData class, which is passed along to [read_csv()](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html).

One reason this might be useful is if a column (like maybe GeoCode) has all numeric values but needs to be treated like a string. For example, if the value `0227` needs to be parsed exactly as `0227` rather than automatically converted to `227.0`, as read_csv normally does for numeric columns.

Usage:

```
  - class: InputCsvData
    path_pattern: data/*.csv
    dtype:
      GeoCode: str
```

### DIFF
--- a/sdg/inputs/InputCsvData.py
+++ b/sdg/inputs/InputCsvData.py
@@ -6,6 +6,19 @@ from sdg.Indicator import Indicator
 class InputCsvData(InputFiles):
     """Sources of SDG data that are local CSV files."""
 
+    def __init__(self, path_pattern='', logging=None,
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin (approved):
Tried this out here: https://lucygwilliamadmin.github.io/open-sdg-simple-starter-1/8-3-1/

Looks good

PR #272, opened 2021-07-26
Fix encoding in constraints CSV

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](add link here)
Related version | 1.5.0-dev
Bugfix, feature or docs? | Bugfix
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

This fixes an issue introduced in #268. That PR used a Pandas parameter that's only available in a specific version of Pandas. By removing some special characters from a CSV file, we don't need that parameter.

### DIFF
--- a/sdg/open_sdg.py
+++ b/sdg/open_sdg.py
@@ -397,6 +397,7 @@ def open_sdg_prep(options):
         # Hardcode some options for global output.
         params['inputs'] = inputs
         params['schema'] = schema
+        params['output_folder'] = options['site_dir']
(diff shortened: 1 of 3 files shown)


brockfanning:
@LucyGwilliamAdmin This is kind of a "hotfix" for #268 - just a couple of issues that weren't quite right in the PR.

LucyGwilliamAdmin (approved):

PR #286, opened 2021-12-01
Csvcubed integration spike

mikeAdamss:
**DON'T MERGE THIS:** Just the easiest way to highlight what I was doing and why.

**Background:**
The SDG team at ONS provides a simple github pages based platform for other government departments to publish their sustainable development goals data.

The way it works roughly is, they provide:

1.) A frontend repo, that each civil service division branches (and can modify with config): https://github.com/open-sdg/open-sdg-site-starter

2.) A data repo, that each civil service division branches, that holds the data that (1) uses: https://github.com/open-sdg/open-sdg-data-starter

3.) A build tool - **the same build tool is used by everyone** and powers creating the static data assets for (2). This repo is a clone of this build tool, this pr is an example of how we could tweak it.

This pr is exploring how you could (potentially, subject to chatting with SDG colleagues ofc) modify that build tool (with csvcubed) so that all those SDG sites are outputting linked data compliant datasets.

**Approach**

The sdg build tool is organised with a suite of outputters (a handler for outputs). All I've done is add an `OutputCsvCube.py` with a new handler and tweaked the constructor to call it.

Seemed more prudent to add a new handler than mess with the existing one (people are rarely against more options, deleting someone else's code might be a little more thorny).

I've also added `localbuild` so you can run the build tool stand along while developing.

**Try it out**

- Clone this branch.
- have `pipenv` installed
- `chmod +x ./localbuild/localbuild.sh`
- `./localbuild/localbuild.sh`

Static assets will appear in `./localbuild/_site/`. If you look in the `/csvw` sub directory one of our style csvws will also be in there.

**TODO**

Broadly, to make this a real thing, we'd need to:

- Figure out what to do about codelists.
- Sort out all the TODO comments I left scattered around.
- Integrate this with the frontend, not a lot of use otherwise.


mikeAdamss:
sorry, not meant to be here at all, ignore.

PR #290, opened 2021-12-22
Translate disaggregation values in geojson output

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](add link here)
Related version | 1.5.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry


brockfanning:
@otis-bath @LucyGwilliamAdmin Any objections to adding this to 2.0.0? This will be needed eventually when map disaggregations are implemented, and should not cause any problems in the meantime. (But of course testing on a FB is a good idea, to ensure maps still work.)

PR #293, opened 2022-02-01
Remove version number from API endpoint

### LINKED ISSUE #291: SDMX API Dataflow version is no longer available
@brockfanning 
Currently for the UN SDMX API input, we are pointing at v1.3 of the SDG Dataflow:
https://github.com/open-sdg/sdg-build/blob/1.8.0-dev/sdg/inputs/InputSdmxMl_UnitedNationsApi.py#L54

However, that is no longer available:
https://data.un.org/ws/rest/data/IAEG-SDGs,DF_SDG_GLH,1.3

We need to update the version (to 1.7?)

This is also something we need to check and implement each version unless there is a way to point at latest version?

brockfanning:
Fixes #291


LucyGwilliamAdmin (approved):
This works :)

PR #294, opened 2022-02-01
Allow multiple data inputs for the same indicator

brockfanning:
@LucyGwilliamAdmin Here is an updated version of this PR. I believe this should get some very critical testing - with both single and multiple data inputs, to ensure there is no unintended consequence.

### DIFF
--- a/sdg/Indicator.py
+++ b/sdg/Indicator.py
@@ -104,7 +104,12 @@ class Indicator(Loggable):
         if val is None or not isinstance(val, pd.DataFrame) or val.empty:
             return
 
-        self.data = val
+        if self.has_data():
+            self.data = pd.concat([self.data, val], sort=False)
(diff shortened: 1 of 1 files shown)


LucyGwilliamAdmin:
@brockfanning once #293 has been merged, is it possible for the latest code to be pulled in here so I can use the global SDMX API as an input please?

brockfanning:
@LucyGwilliamAdmin Yes, that's done.

LucyGwilliamAdmin:
Testing checklist
[ ] 1 source
[ ] 2 source - different series coming from two sources
[ ] 2 source - data for same series coming from two sources e.g. different disaggs

LucyGwilliamAdmin:
@brockfanning I've added some scenarios that I'm testing above - can you think of any others?

PR #295, opened 2022-02-01
Special treatment of more SDMX-related columns

brockfanning:
This may be needed for special footer treatment of some SDMX columns in Open SDG.


brockfanning:
@otis-bath @LucyGwilliamAdmin I think this should go in with 1.8.0. It's not required, but it would make sense.

otis-bath (approved):

PR #316, opened 2022-06-24
Rename columns that become duplicates after translation

### LINKED ISSUE #315: Issue with multiple data inputs
When I try to input from SDMX and CSV together, this works fine as long as `drop_singleton_dimensions` is true. Similarly if I only have SDMX input and `drop_singleton_dimensions` is false, this also works.

So the specific case when an the issue occurs is two data inputs (SDMX and CSV) and `drop_singleton_dimensions` is false.

Error: https://github.com/uk-sdmx-api-platform/data/runs/7039043267?check_suite_focus=true#step:5:1252 - the data is being printed and there's 2 dfs for 1.1.1 so not sure if this could be something to do with it, since the error refers to a duplicate axis.

Data config file: https://github.com/uk-sdmx-api-platform/data/blob/develop/config_data.yml#L42-L50

brockfanning:
Q | A
--- | ---
Feature branch/test site URL | [Link](https://uk-sdmx-api-platform.github.io/site/1-1-1/)
Fixed issues | Fixes #315 
Related version | 2.0.0-dev
Bugfix, feature or docs? |
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry


LucyGwilliamAdmin:
I've just added link to description - I think this is what's expected right?

brockfanning:
@LucyGwilliamAdmin Yes, that seems like what it's coded to do.

LucyGwilliamAdmin (approved):
Works as expected

PR #322, opened 2022-08-20
Handle Pandas bug in Series.replace

### LINKED ISSUE #321: Error when using code_map in inputs
Pandas 1.4 introduced a breaking change in the [Series.replace](https://pandas.pydata.org/docs/reference/api/pandas.Series.replace.html) method. If you are using Pandas 1.4 or higher and try to use the code_map parameter, you'll see this error: `ValueError: Series.replace cannot use dict-like to_replace and non-None value`. More info on the bug here: https://github.com/pandas-dev/pandas/issues/46004

I'm not sure if Pandas is going to revert their breaking change (they should IMO, since it was introduced in a minor version), but we need to adjust our code in the meantime. The work-around in the meantime is to not specify the `value` parameter at all. But, we still need it for previous versions of Pandas... So we will need to do a try/except and use both.

brockfanning:
Fixes #321 

Testing instructions: @otis-bath @sarah-beasley The reason that this bug has not been a problem for the UK data builds is that you are using Python 3.7 right now. To see this bug you could update your version of Python to 3.8. To do that in a feature branch you would need to update [this line](https://github.com/ONSdigital/sdg-data/blob/develop/.github/workflows/deploy-feature-branch.yml#L29). I believe that if you try that, the build will fail. Then, if you point to this branch in your requirements.txt file, the build should succeed.

So in sum, this bug affects any site using Python 3.8 and also using the `code_map` parameter in the data config. The UK build is indeed using the `code_map` parameter in the `sdmx_output` section.


otis-bath (approved):
This is working as intended

PR #323, opened 2022-08-21
Include class name in context for alterations

brockfanning:
Data and metadata alterations are applied to particular inputs, but in Open SDG there are often many inputs. For example there might be a CSV data input, an SDMX metadata input, a YAML indicator configuration input, etc. By including the class name of the input in the context object, we can give people a more robust way to target a particular input.


brockfanning:
This change is a harmless addition and does not need manual testing, so I'll go ahead and merge based on the passing tests.

PR #324, opened 2022-08-21
Alter indicators knowing data and metadata

brockfanning:
A commonly requested feature in Open SDG has been the ability to alter data with a custom Python function that has access to the metadata. Or vice versa, to alter metadata with a custom Python function that has access to the data. This has been difficult because our alteration functions apply to each input, and Open SDG uses many inputs, most of which are only for data or only for metadata.

This adds an "alter_indicator" option that is applied to the *outputs* (not the inputs) after the data and metadata have been completed merged together from all the inputs. The thing being altered is the Indicator object, which has data and metadata in it.

The existing "alter_meta" and "alter_data" options will continue to work the same as they have.

Example of usage:

```
from sdg.open_sdg import open_sdg_build

def my_indicator_alteration(indicator, context):
    # If the number of years is 2 or less, use a bar chart.
    if indicator.data.Year.nunique() <= 2:
        indicator.meta['graph_type'] = 'bar'
    return indicator

open_sdg_build(config='config_data.yml', alter_indicator=my_indicator_alteration)
```


LucyGwilliamAdmin:
@brockfanning tested this and looks to be working: http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-sdg-build-324

PR #330, opened 2022-09-22
Better error message when CSV whitespace issues

brockfanning:


sarah-beasley:
Hi @brockfanning I've been trying to test these but have been coming across some problems. The first being that when adding a trailing white space before pointing to your feature branch, there was no error that seemed to come up. But now pointing to your feature branch in the requirements.txt file, [this](https://github.com/ONSdigital/sdg-data/actions/runs/4344594262/jobs/7588126209) error comes up, and this is without there being any trailing white spaces added either.

brockfanning:
@sarah-beasley This PR is different from most because I accidentally pushed it to the origin instead of my fork. So instead of the usual "brockfanng/sdg-build-1" it will actually be "open-sdg/sdg-build". Eg: `git+https://github.com/open-sdg/sdg-build@brockfanning-patch-2`

About the other issue though - I think the problem might be that your deploy-feature-branch workflow doesn't run `python scripts/check_data.py`. (It only runs `python scripts/build_data.py`.) There are a couple of ways to work around this issue:

1. We could update your deploy-feature-branch workflow so that it does run `python scripts/check_data.py`. The only downside there is that it means that deploying feature branches will take a few minutes longer, because I believe the `check_data.py` script takes a few minutes to run.
2. You could create a pull-request using your feature branch, which would trigger the test-pull-requests workflow, which does run `check_data.py`.

sarah-beasley:
Thanks @brockfanning that was really helpful. I raised a pull request and the error message now shows as this which is great and much more informative for people to fix:
image


Just one other thing, I tested the FB created before raising the pull request to see if there were any changes and where I added a space after percentage for the first two fields of 10-4-1, it has made them into another series. I guess you wouldn't be able to merge without fixing due to the tests failing but would you expect this to happen? I guess because it thinks there are two types of unit now - one with a space and one without?:
image

brockfanning:
@sarah-beasley I guess that behavior that you're seeing (the double unit) is one of the reasons that the test is important. That does seem expected though, because -- to a computer -- the added space does indeed make it a different unit. There is code we could write to automatically remove extra spaces, but I would argue that the test we already have is a better way to catch it, because it alerts the user that they did something wrong, and encourages attention to detail.

PR #333, opened 2023-01-14
Prevent version 5 of frictionless

brockfanning:
Frictionless v5 requires Python 3.8, and we have not yet confirmed compatibility with Python 3.8. We can remove this restriction once we're sure that using Python 3.8 is fine.

Q | A
--- | ---
Feature branch/test site URL | [Link](http://uk-sdg-feature-branches.s3-website.eu-west-2.amazonaws.com/feature-stop-pin-frictionless/)
Related version | 2.2.0-dev
Bugfix, feature or docs? | Bugfix
* [ ] Includes tests
* [ ] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

In theory, with this PR, countries would not need to "pin" the Frictionless package in their requirements.txt file. So this should be tested along with removing that from the requirements.txt file (the line `frictionless==4.40.8` could be remove)


otis-bath (approved):
@brockfanning This has built all okay and the FB looks good

PR #336, opened 2023-03-05
Halt build and provide better error messages on YAML syntax problems

brockfanning:
Fixes https://github.com/open-sdg/open-sdg/issues/1766

Testing instructions:

This PR includes the following changes:

1. YAML syntax problems in metadata files and indicator configuration files will receive a more helpful error message.
2. YAML syntax problems in translation files will receive a more helpful error message, **and will now prevent the build from continuing**.

I bolded part of the second item above because it is kind of a "breaking change" in that it may prevent builds from completing where previously the completed. However it could be argued that this breaking change is OK, because previously it was failing silently to import the translation files, which probably is causing confusion. So now it will fail, but will remove confusion.

Here is an example of the syntax problem this PR should now catch:

```
foo: 'hello 'open' sdg'
```

To test this, you can try adding the line above in any/all of:

1. YAML indicator metadata files
2. YAML indicator configuration files
3. YAML translation files

### DIFF
--- a/sdg/helpers/files.py
+++ b/sdg/helpers/files.py
@@ -23,3 +23,11 @@ def read_file(location, request_params=None):
         data = file.read()
     file.close()
     return data
+
+
+def print_yaml_syntax_help(file):
+    print('-----')
+    print('The file at ' + file + ' could not be parsed because of a syntax error.')
+    print('YAML syntax errors often involve single/double quotes and/or colons (:).')
+    print('Sometimes you can find the problem by looking at the lines/columns mentioned in the following raw error message:')
+    print('------')
--- a/sdg/helpers/files.py
+++ b/sdg/helpers/files.py
@@ -25,9 +25,9 @@ def read_file(location, request_params=None):
     return data
 
 
-def print_yaml_syntax_help(file):
-    print('-----')
-    print(f'The file at {file} could not be parsed because of a syntax error.')
-    print('YAML syntax errors often involve single/double quotes and/or colons (:).')
-    print('Sometimes you can find the problem by looking at the lines/columns mentioned in the following raw error message:')
-    print('------')
+def get_yaml_syntax_help(file):
+    return f"""-----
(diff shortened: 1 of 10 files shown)


otis-bath:
@brockfanning since we don't use yml for our metadata/indicator config is there a way we are able to test this?

brockfanning:
@otis-bath You could try removing the `translations` section from your config_data.yml file, and then adding this:

```
inputs:
  - class: InputCsvData
    path_pattern: data/*.csv
  - class: InputYamlMeta
    path_pattern: indicator-config/*.yml
    git: false
  - class: InputYamlMdMeta
    path_pattern: meta/*.md
    git: true
    git_data_dir: data

translations:
    source: https://github.com/open-sdg/sdg-translations.git
    tag: 2.1.0
  - class: TranslationInputYaml
    source: translations
```

Then you could create two new files: `indicator-config/1-1-1.yml` and `translations/en/foo.yml`. The contents of both could the same snippet as above:

```
foo: 'hello 'open' sdg'
```

brockfanning:
@otis-bath It's there, but you have to scroll up a bit. There are a few annoying warnings in there (`warn(f"Print area cannot be set to Defined name: {defn.value}.")`) that I would like to get rid of. Let me see if I can get rid of those.

brockfanning:
@otis-bath Could you trigger it again? I've made a slight tweak to hopefully avoid those warnings.

otis-bath:
@brockfanning refreshed - you can see the readout here https://github.com/ONSdigital/sdg-data/actions/runs/4345850379/jobs/7591124715

brockfanning:
@otis-bath Thanks - I believe I've got it now so that the error message should show up at the very bottom. Could you trigger it again?

otis-bath (approved):
Thanks Brock - this is working and error message is now appearing at the bottom:
https://github.com/ONSdigital/sdg-data/actions/runs/4356330984/jobs/7614187094

PR #337, opened 2023-03-16
Avoid key error when report field does not exist in indicator metadata

brockfanning:
Testing instructions:

This avoids an error that happens when the metadata report is used with a field that does not have any metadata. For example, the following data config:

```
docs_metadata_fields:
    - key: foo
      label: Foo
```

That would cause a build error, assuming no indicators actually have "foo" metadata. But with this PR, the build would succeed and the metadata report would have a "Foo" column that is blank.


otis-bath (approved):
Thanks @brockfanning - can confirm the tests pass when pointing at the FB but don't when pointed at 2.1.0 of sdg-build
image

PR #339, opened 2023-04-05
Pandas version - pin to 1.x

brockfanning:
The latest version of Pandas (2.0.0) isn't compatible with something in Frictionless, so let's prevent this library from using Pandas 2.0.0, for now.


brockfanning:
@sarah-beasley Tests are failing because of an issue with the UN API, but can you go ahead and try this out?

sarah-beasley:
Not sure if it's a related problem but I can't seem to get the site FB to build! I've tried by just changing the config.yml in the site repo to point to your branch, and also pointing to your branch in the data repo and then the site repo, but both errored the same at the site stage - https://github.com/ONSdigital/sdg-indicators/actions/runs/4627515638/jobs/8185518573. Sorry if it's just something I have done wrong

brockfanning:
@sarah-beasley Because this is an sdg-build PR, you would point to this branch in the requirements.txt field of the data repo, like this: `git+https://github.com/brockfanning/sdg-build-1@pandas-version`

brockfanning:
@sarah-beasley And to clarify, you would not need to point to this branch in the site repo -- only the data repo uses sdg-build.

brockfanning:
Thanks @sarah-beasley. I think the UN API may be blocking Github, and that's why the tests are failing. So I think it's OK to merge this one.

PR #359, opened 2024-11-05
Progress measure

tristanmenard:
Q | A
--- | ---
Feature branch/test site URL | https://tristanmenard.github.io/site/
Fixed issues | 
Related version | 2.4.0-dev
Bugfix, feature or docs? | Feature
* [x] Includes tests
* [x] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry

Changes from the previous PR:
* Re-organized everything into classes built upon the IndicatorProgress class.
* Corrected math for cases where the desired direction of progress differs from the direction to the target. When the initial value achieves the target, the desired direction of progress and the direction to the target are opposite, which previously led to incorrect results.
* Corrected math for cases where values are < 0 (invert the desired direction of progress when values are < 0)
* Added the ability to choose which series and/or units to use for the progress calculations in the metadata files. If there is no headline for the indicator, the user can also select disaggregations. Some examples below.
  * To specify the series and unit for an indicator with a headline: 
  ```
  progress_calculation_options:
  - series: Series1
    unit: Unit1
  ```
  * To specify the series/unit and disaggregation for an indicator without a headline:
  ```
  progress_calculation_options:
  - series: Series1
    unit: Unit1
    disaggregation:
    - field: Geography
      value: Canada
    - field: Gender
      value: Women
  ```
  * It is possible to choose multiple series/units/disaggregations:
  ```
  progress_calculation_options:
  - series: Series1
    unit: Unit1
    disaggregation:
    - field: Geography
      value: Canada
    - field: Gender
      value: Women
  - series: Series2
    unit: Unit1
  ```
* Enabled progress calculations for indicators with no headline but many sub-indicator series. The ~~minimum~~ **mean** progress score among the series is taken as the overall progress score for the indicator.
* Implemented a scaling factor to reduce the fixed progress thresholds when the initial value is near some natural limit for the indicator (for example if an indicator cannot exceed 100%) for indicators with qualitative targets.
* Added a progress column that can be used to override the values in the progress calculations. Allows for data that is inappropriate for the progress measure calculation to be converted into a form that is appropriate.


tristanmenard:
@brockfanning Currently, the progress calculation is done inside OutputOpenSdg, but it seems to be running the calculation for each language. Ideally, the progress calculation could be done only once. Any ideas where to move the call to IndicatorProgress so that it doesn't redo the calculations for every language?

PR #362, opened 2024-12-06
More helpful message when SDMX input fails.

brockfanning:


brockfanning:
@Govej Here's the more helpful error message we talked about today.

PR #369, opened 2025-05-27
Progress measure deepcopy and not_available logic

tristanmenard:
Q | A
--- | ---
Feature branch/test site URL | https://tristanmenard.github.io/scratch-site/
Fixed issues | not_available progress status logic AND year truncation in chart when auto_progress_calculation turned on
Related version | 2.4.0-dev
Bugfix, feature or docs? | bugfix
* [x] Includes tests
* [x] Updated docs
* [ ] Updated config forms
* [ ] Added CHANGELOG entry


tristanmenard:
I just added the ability to group series together so that the mean score of the grouped series is used in the calculation of the aggregate score for the indicator. This means that for an indicator like 1.4.1, we can take the equi-weighted mean of the scores from each form of basic access (even though there may be different numbers of sub-series that contribute towards each form of basic access).

The way I chose to format groupings in the progress_calculation_options is like this:
```
progress_calculation_options:
- group:
  - series: Access to basic education 1
    direction: positive
  - series: Access to basic education 2
    direction: positive
- series: Access to basic health services
  direction: positive
```
where the education series are grouped and the health services series is on its own. This means that the final aggregate score is calculated as `mean([mean([Access to basic education 1, Access to basic education 2]), Access to basic health services])`.

@brockfanning I'm not sure if this is easy to implement in the forms version or not? Let me know if the formatting should be changed.

Of course, it's fully backwards compatible with the previous score calculation method.

For context, I was trying to add a weighting method which I thought would be fairly simple but I had to switch to grouping when I realized it would be complicated to figure out how to re-distribute weights of series with scores of None.

